PR #2, opened 2015-06-25
Add app user

GordonDiggs:
/cc @codeclimate/review 

@pbrisbin is this what you had in mind for the way engines should add the user?


pbrisbin:
This LGTM, though I'm no docker expert.

PR #3, opened 2015-06-30
Handle alternate location class

GordonDiggs:
/cc @codeclimate/review 

The Syntax check returns a different location object than all the other checks

### DIFF
--- a/lib/cc/engine/rubocop.rb
+++ b/lib/cc/engine/rubocop.rb
@@ -64,6 +64,31 @@ module CC
         RuboCop::Cop::Team.new(RuboCop::Cop::Cop.all, rubocop_config)
       end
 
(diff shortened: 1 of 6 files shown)


lib/cc/engine/rubocop.rb
```diff
@@ -64,6 +64,31 @@ def rubocop_team_for_path(path)
         RuboCop::Cop::Team.new(RuboCop::Cop::Cop.all, rubocop_config)
       end
 
+      def violation_positions(location)
+        if location.is_a? RuboCop::Cop::Lint::Syntax::PseudoSourceRange
```
pbrisbin: :lipstick: invoke with parens

pbrisbin:
LGTM

PR #4, opened 2015-07-02
Read config from json file

GordonDiggs:
/cc @codeclimate/review 

I also switched this up to match other engines (accept the hash of the config, rather than the json string)

### DIFF
--- a/spec/cc/engine/rubocop_spec.rb
+++ b/spec/cc/engine/rubocop_spec.rb
@@ -85,7 +85,7 @@ module CC::Engine
             return false
           end
         EORUBY
-        config = {"exclude_paths" => ['my_script']}
+        config = { "exclude_paths" => ["my_script"] }
         output = run_engine(config)
         assert !includes_check?(output, "Lint/UselessAssignment")
       end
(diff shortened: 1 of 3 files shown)


spec/cc/engine/rubocop_spec.rb
```diff
@@ -85,7 +85,7 @@ def method
             return false
           end
         EORUBY
-        config = {exclude_paths: ['my_script']}.to_json
+        config = {"exclude_paths" => ['my_script']}
```
pbrisbin: :lipstick: `{ ` and ` }`

spec/cc/engine/rubocop_spec.rb
```diff
@@ -107,7 +107,7 @@ def method
           "rubocop.yml",
           "AllCops:\n  Exclude:\n    - \"foo.rb\"\n"
         )
-        config = {config: 'rubocop.yml', exclude_paths: ['bar.rb']}.to_json
+        config = { "config" => 'rubocop.yml', "exclude_paths" => ['bar.rb'] }
```
pbrisbin: :lipstick: double quote hash values

pbrisbin:
Some existing style things worth fixing while here (I think) -- LGTM

PR #6, opened 2015-07-09
Improve tests around rubocop.yml config files

danielmorrison:
I was having issues and discovered that one of these test was a false positive. I then added another test to confirm that `.rubocop.yml` is used without specifying it in the config.


pbrisbin:
Thanks! I'm going to spend a few minutes looking into the Circle issue, but will likely merge this anyway.

/cc @GordonDiggs

PR #9, opened 2015-08-18
WIP: cyclomatic complexity: indicate full method in location

wfleming:
I believe that [this trello card](https://trello.com/c/XAEGRo0B/701-rubocop-cyclomatic-complexity-method-does-not-allow-expansion-view-more-of-source-listing) is because RuboCop's behavior is to return only the method definition line for all complexity checkers.

The fix here is dirty, but the alternative seems kind of worse. Commit message on 00b63ed has more details on how RuboCop behaves, and why monkey patching seems better in this case.

Thoughts, @codeclimate/review?


lib/rubocop/cop/method_complexity_patch.rb
```diff
@@ -0,0 +1,10 @@
+# Monkey patching! Here be dragons.
+module RuboCop
+  module Cop
+    module MethodComplexity
+      def add_offense(node, loc, message = nil, severity = nil)
```
GordonDiggs: What do you think of adding a comment here about what/why this is? Similar to what you have in the commit message

wfleming: Yeah, thanks, I agree. Want to keep it shorter than the commit message explanation, and not sure how I want to word it yet. I'll ping you when I have something.

pbrisbin: FWIW: I'm a big fan of good commit messages and a team that goes to `git blame` immediately over leaving "wtf?" comments in the source. In this case a small comment and big commit message seems like an OK compromise though.

wfleming: Well, I've added a comment now, but it's not short. I'm not sure how much briefer it can be & still be good english without just cutting it down to "see commit message for why this is begin done."

pbrisbin: I don't feel strongly, I think this is fine as-is.

GordonDiggs:
LGTM

lib/rubocop/cop/method_complexity_patch.rb
```diff
@@ -0,0 +1,18 @@
+# Monkey patching! Here be dragons.
+module RuboCop
+  module Cop
+    module MethodComplexity
+      # RuboCop's default implementation of `add_offense` in `Cop` only gets the
+      # location of the keyword associated with the problem which, for things
+      # like complexity checkers, is just the method def line. This isn't very
+      # useful for checkers where the entire method body is relevant. Fetching
+      # this information from an `Offense` instance is difficult, since the
+      # original AST is no longer available. So it's easier to monkey-path
```
ABaldwinHunter: `monkey-patch` :star: 

I like the detail in this explanation. If you prefer brevity, what about something like:

`Overwrite RuboCop to send entire offending method as offense, not just def line.`

lib/rubocop/cop/method_complexity_patch.rb
```diff
@@ -0,0 +1,18 @@
+# Monkey patching! Here be dragons.
+module RuboCop
+  module Cop
+    module MethodComplexity
```
brynary: I'd define this as:

```
module RuboCop::Cop
  MethodComplexity.module_eval do
    # ...
  end
end
```

Reason: If any of these module names change, it will fail fast.

brynary: Or:

```
RuboCop::Cop::MethodComplexity.module_eval do
end
```

brynary:
@wfleming @GordonDiggs we should open an upstream ticket to check if this is something that would be valuable to Rubocop proper. It seems like it would be better to keep the whole node location info, and then if e.g. a text-based formatter just wants to print a single line, it can.

PR #10, opened 2015-08-20
change monkey patch to use module_eval

wfleming:
Bryan [suggested this](https://github.com/codeclimate/codeclimate-rubocop/pull/9#discussion-diff-37535714R4) as a refactor as it will fail faster if RuboCop changes in the future, and it will fail in a more obvious way (rather than just suddenly failing one of our analysis specs). Seems sound to me.

This won't affect current behavior, just guard a bit against future failure, so there's no rush in merging or deploying.


GordonDiggs:
LGTM

PR #11, opened 2015-08-25
update badge link

ABaldwinHunter:
@codeclimate/review Updates badge link

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:edge
+FROM codeclimate/alpine-ruby:b38
 
 WORKDIR /usr/src/app
 COPY Gemfile /usr/src/app/
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Code Climate Rubocop Engine
 
-[![Code Climate](https://codeclimate.com/repos/5584174c6956807f6701c8f5/badges/9b8937da146a0abf21d4/gpa.svg)](https://codeclimate.com/repos/5584174c6956807f6701c8f5/feed)
+[![Code Climate](https://codeclimate.com/repos/55ccc68e6956802d030012f3/badges/fce009562bca1c1fc8dc/gpa.svg)](https://codeclimate.com/repos/55ccc68e6956802d030012f3/feed)
 
 `codeclimate-rubocop` is a Code Climate engine that wraps the [RuboCop](https://github.com/bbatsov/rubocop) static analysis tool. You can run it on your command line using the Code Climate CLI, or on our hosted analysis platform.
 


pbrisbin:
Looks like this is hitting the same error @mrb was...

wfleming:
`gem install io-console` in the Dockerfile was his fix. I still think a root cause must be something with the alpine ruby image.

ABaldwinHunter:
@pbrisbin @wfleming For this issue, I can try to do `gem install io-console`.

Or, do we want to use our own copy of `alpine-ruby` stored somewhere?

pbrisbin:
I think it makes sense to move to our alpine-ruby base image if that's what we do for all our other ruby-based engines. I don't feel strongly though, and moving to some other defacto ruby base image is also fine with me.

ABaldwinHunter:
@codeclimate/review I updated the `Dockerfile` to use `codeclimate/alpine-ruby:b36` instead of `alpine:edge` and the tests pass. 

I'm also okay waiting on this change.

Eventually, it might be preferable to use `ruby:2.2.2` but that will involve a more in depth change to the way we run `app` user.

pbrisbin:
LGTM

Dockerfile
```diff
@@ -1,4 +1,8 @@
-FROM alpine:edge
+FROM codeclimate/alpine-ruby:b36
+
+WORKDIR /usr/src/app
```
GordonDiggs: These lines are duplicates

Dockerfile
```diff
@@ -1,4 +1,8 @@
-FROM alpine:edge
+FROM codeclimate/alpine-ruby:b36
```
GordonDiggs: Should this be 38?

ABaldwinHunter: I believe we only have `b36` publicly atm https://hub.docker.com/r/codeclimate/alpine-ruby/tags/

```
$ docker pull codeclimate/alpine-ruby:b38
Pulling repository codeclimate/alpine-ruby
FATA[0001] Tag b38 not found in repository codeclimate/alpine-ruby 
```

Dockerfile
ABaldwinHunter: thanks!

GordonDiggs:
LGTM

ABaldwinHunter:
@codeclimate/review `codeclimate/alpine-ruby:b38` is now available on Dockerhub (thanks @GordonDiggs!) so I opted to use that image instead. (It's used by `scheduler`). 

Using this new `rubocop`, I successfully ran analysis locally via CLI `codeclimate analyze` and with SOA of `harness`.

PR #12, opened 2015-08-28
update circle.yml to tag git & build to docker hub

wfleming:
cc @codeclimate/review: this is a proposal for a model that we could apply to all our engines.

* tag all successful master builds in git (so it's easy to correlate a
  commit with a docker image tag)
* push built docker images to both our private repository &  the public
  docker repository (which is used by the `codeclimate` cli)
* Adding the `owner` field to the `deploy` should make the experience a bit nicer for open source contributors who want to work on their own fork: they can commit to their own master without triggering our deploy behavior (which, of course, wouldn't work for them).

We could also set the ENV for the docker hub URL right here in the yml, or just hardcode it. I think there was some discussion elsewhere of moving away from managing those values in circle, where feasible.


GordonDiggs:
:+1: I like it.

circle.yml
```diff
@@ -15,11 +14,14 @@ dependencies:
...
 
 deployment:
   registry:
     branch: master
+    owner: codeclimate
     commands:
-      - docker push $registry_root/$image_name:b$CIRCLE_BUILD_NUM
+      - git tag b$CIRCLE_BUILD_NUM master && git push origin b$CIRCLE_BUILD_NUM
+      - docker push $registry_root/$CIRCLE_PROJECT_REPONAME:b$CIRCLE_BUILD_NUM
+      - docker push $public_registry_root/$CIRCLE_PROJECT_REPONAME:b$CIRCLE_BUILD_NUM
```
pbrisbin: I think you need to `docker tag $registry_root/... $public_registry_root/...` before this push will work.

circle.yml
```diff
@@ -15,11 +14,14 @@ dependencies:
 test:
   override:
     - bundle exec rake
-    - docker build -t=$registry_root/$image_name:b$CIRCLE_BUILD_NUM .
+    - docker build -t=$registry_root/$CIRCLE_PROJECT_REPONAME:b$CIRCLE_BUILD_NUM .
```
pbrisbin: Any variables used here will have to be set in the circle.yml or it will fail on contributor's PRs (since it'll be unset for those builds). That's the reason it was moved in the CLI repo.

I'd also suggest renaming to `REGISTRY_ROOT` (and `PUBLIC_REGISTRY_ROOT`) since environment variables are conventionally capitalized.

wfleming: Makes sense to me: I'll compare to CLI & update. Definitely agree about capitalization: I wasn't sure if there was a reason they weren't already that way.

circle.yml
```diff
@@ -3,23 +3,26 @@ machine:
...
 deployment:
   registry:
     branch: master
+    owner: codeclimate
     commands:
-      - docker push $registry_root/$image_name:b$CIRCLE_BUILD_NUM
-
+     - echo $GCLOUD_JSON_KEY_BASE64 | sed 's/ //g' | base64 -d > /tmp/gcloud_key.json
+      - curl https://sdk.cloud.google.com | bash
+      - gcloud auth activate-service-account $GCLOUD_ACCOUNT_EMAIL --key-file /tmp/gcloud_key.json
```
wfleming: Comparing to https://github.com/codeclimate/codeclimate/blob/master/circle.yml, I notice we don't specify the email there. Do we need to generally & don't there for some reason?

circle.yml
```diff
@@ -3,23 +3,26 @@ machine:
...
+      - gcloud docker -a
+      - git tag b$CIRCLE_BUILD_NUM master && git push origin b$CIRCLE_BUILD_NUM
+      - docker push $PRIVATE_REGISTRY/$CIRCLE_PROJECT_REPONAME:b$CIRCLE_BUILD_NUM
+      - docker tag $PRIVATE_REGISTRY/$CIRCLE_PROJECT_REPONAME:b$CIRCLE_BUILD_NUM $PUBLIC_REGISTRY/$CIRCLE_PROJECT_REPONAME:b$CIRCLE_BUILD_NUM
+      - docker push $PUBLIC_REGISTRY/$CIRCLE_PROJECT_REPONAME:b$CIRCLE_BUILD_NUM
+      - docker tag -f $PUBLIC_REGISTRY/$CIRCLE_PROJECT_REPONAME:b$CIRCLE_BUILD_NUM $PUBLIC_REGISTRY/$CIRCLE_PROJECT_REPONAME:latest
```
wfleming: I realized I was pushing the tag but not `latest` for the public docker hub registry. Since that's what the CLI actually uses, I think we need to push both?

circle.yml
```diff
@@ -3,23 +3,26 @@ machine:
...
 deployment:
   registry:
     branch: master
+    owner: codeclimate
     commands:
-      - docker push $registry_root/$image_name:b$CIRCLE_BUILD_NUM
-
+      - echo $GCLOUD_JSON_KEY_BASE64 | sed 's/ //g' | base64 -d > /tmp/gcloud_key.json
+      - curl https://sdk.cloud.google.com | bash
+      - gcloud auth activate-service-account $GCLOUD_ACCOUNT_EMAIL --key-file /tmp/gcloud_key.json
```
wfleming: Comparing to https://github.com/codeclimate/codeclimate/blob/master/circle.yml, I notice we don't specify the email there. Do we need to generally & don't there for some reason, or can we drop this?

pbrisbin: I'd consider the key and email as both secret (even though the email's not really sensitive) and prefer to keep them in the Circle project settings. The others are both not sensitive and *must* be hardcoded for contributor PRs to work (because they're used in the `test` step).

circle.yml
pbrisbin: I think updating `:latest`should be an explicit step, done manually. Unless the team feels it's OK to start effectively auto-deploying any engine to CLI users when commits land in master.

circle.yml
wfleming: I'm saying the email doesn't appear in codeclimate/codeclimate's env *at all* AFAICT. It's not specified in the ENV on circleci, and it's not used in the `gloud auth` command there. So I'm wondering if the private key is in and of itself sufficient (it seems like it is, if the codeclimate/codeclimate yml is valid) and, if so, if we think that's what we should do in general.

circle.yml
wfleming: Fair point. Maybe "tag latest" should be a fab/shipbot task. I don't think the CLI yet automatically pulls updated images, though, if that makes a difference to how we feel about it.

wfleming: CLI not auto-updating actually makes this worse, I guess: if somebody happened to install & start using it when we had a bad engine image for any reason, they'd keep having the bad engine version until they did an explicit pull, and wouldn't even necessarily know that a pull was necessary. So I'll delete for now.

Once CLI auto-updates itself & images, though, I think this might be tolerable & worth it: if an image were bad for any reason, we'll have to notice and repair or rollback quickly in the cloud, and making it automatic makes in one less thing we need to remember to do manually, so it would stop CLI engines from getting too far behind the cloud from forgetfulness.

circle.yml
pbrisbin: Oh sure. That's not how I remembered it, but if it works without it then even better

pbrisbin:
LGTM. Won't know if it works until you merge anyway.

PR #13, opened 2015-08-31
Fix issue with DisabledByDefault in rubocop config file 

jacobi007:
It should return 0 issue if this option is set to true

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,22 +2,22 @@ GEM
   remote: https://rubygems.org/
   specs:
     ansi (1.5.0)
-    ast (2.0.0)
-    astrolabe (1.3.0)
-      parser (>= 2.2.0.pre.3, < 3.0)
+    ast (2.1.0)
+    astrolabe (1.3.1)
+      parser (~> 2.2)
     builder (3.2.2)
     coderay (1.1.0)
     metaclass (0.0.4)
     method_source (0.8.2)
-    minitest (5.6.0)
-    minitest-reporters (1.0.11)
+    minitest (5.8.0)
+    minitest-reporters (1.0.20)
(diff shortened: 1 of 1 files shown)


Gemfile.lock
```diff
@@ -26,9 +26,9 @@ GEM
       slop (~> 3.4)
     rainbow (2.0.0)
     rake (10.4.2)
-    rubocop (0.31.0)
+    rubocop (0.33.0)
```
pbrisbin: Is this update intentional? Does it fix or change some aspect of config handling?

jacobi007: Yes, update was required because the `DisabledByDefault` option was added in version 0.33 - https://github.com/bbatsov/rubocop/blob/master/CHANGELOG.md#new-features-1

lib/cc/engine/rubocop.rb
```diff
@@ -45,7 +45,10 @@ def exclude?(local_path)
       def rubocop_config_store
         @rubocop_config_store ||= begin
           config_store = RuboCop::ConfigStore.new
-          if (config_file = @engine_config["config"])
+
+          if (default_rubocop_config = Dir.glob("/code/.rubocop.yml").first)
```
jacobi007: You are 100% correct, I've just checked and this modification is not necessary. I will revert it in a next commit.

mrb:
@pbrisbin Looks like this was updated to just include the Gem change - do you :+1:?

pbrisbin:
LGTM.

If you could squash to one commit with a good message that'd be aces!

> Upgrade rubocop gem
>
> - vWhatever brings DisabledByDefault support
> - link to github compare of old/new version

Thanks!

PR #14, opened 2015-08-31
circleci: don't tag to git or deploy to docker hub

wfleming:
Doing all of this in Circle turned out to be a pain to get working correctly, and would have been even more of a pain to expand to other repos, so I'm backing out those bits.

Eventually, it would be good to have all of those various tasks centralized: once we do that elsewhere, we can remove the `deployment` bit entirely here. In the meantime, we should keep the existing behavior.


pbrisbin:
LGTM. Want to also add the notifications hook while you're in here?

pbrisbin:
https://github.com/codeclimate/codeclimate-ops/blob/master/circle.yml#L25-L27

PR #16, opened 2015-09-01
Add content to issue JSON

grstk:
TODO: add actual contents for ClassLength, ModuleLength and AbcSize checks.

### DIFF
--- a/lib/cc/engine/rubocop.rb
+++ b/lib/cc/engine/rubocop.rb
@@ -3,6 +3,8 @@ require "pathname"
 require "rubocop"
 require "rubocop/cop/method_complexity_patch"
 require "cc/engine/category_parser"
+require "active_support"
+require "active_support/core_ext"
 
 module CC
   module Engine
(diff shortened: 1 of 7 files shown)


mrb:
Needs review please /cc @codeclimate/review

lib/cc/engine/rubocop.rb
```diff
@@ -83,17 +85,25 @@ def violation_positions(location)
...
+                            type: "Issue",
+                            check_name: "Rubocop/#{violation.cop_name}",
+                            description: violation.message,
+                            categories: [category(violation.cop_name)],
+                            remediation_points: 50_000,
+                            location: {
+                              path: local_path,
+                              positions: violation_positions(violation.location),
+                            },
+                         }
```
pbrisbin: Can you format this as:

```rb
violation = {
  type: ...,
  check_name: ...,
}
```

That's our style and will make the diff clearer too.

spec/cc/engine/rubocop_spec.rb
```diff
@@ -201,10 +201,27 @@ def method(a,b,c,d,e,f,g)
         assert_equal location, result["location"]
       end
 
+      it 'includes issue content when available' do
```
pbrisbin: :lipstick: please use double quotes throughout.

pbrisbin:
Code LGTM (modulo some minor nits). I don't think we can merge until the TODO is addressed though.

GordonDiggs:
Is the plan that we're going to copy and paste in content for every cop? The classes define [their own descriptions](http://www.rubydoc.info/gems/rubocop/0.27.0/RuboCop/Cop/Metrics/AbcSize), though it is as a comment in the code. Is there a way we could interface with that? Perhaps by parsing the YARD documentation and accessing it that way?

GordonDiggs:
Copying and pasting doesn't seem like a very sustainable solution, is all

grstk:
@GordonDiggs I don't think the content for check is supposed to be the same as cop definition from comments (I was told that the contents for the three checks above were 'previously written'). @brynary can you confirm?

pbrisbin:
He probably mean't to bring in the existing ReadUp content for the similar complexity checks we have in non-platform analysis.

@mrb can you confirm? If so, would you be able to find the correct content in admin/content_blobs and copy them in here?

brynary:
What @pbrisbin said is correct.

mrb:
Yep I got it @pbrisbin

PR #19, opened 2015-09-17
We're hiring!

wfleming:
@noahd1 pretty sure this is the file Bryan was referring to. Want me to drop it into all the open source engines?


noahd1:
:+1: yes - please, all CC owned repos

PR #20, opened 2015-09-18
Removes non-Ruby files from include_path before passing to rubocop.

fhwang:
cc @codeclimate/review

### DIFF
--- a/lib/cc/engine/file_list_resolver.rb
+++ b/lib/cc/engine/file_list_resolver.rb
@@ -1,0 +1,61 @@
+module CC
+  module Engine
+    class FileListResolver
+      def initialize(code:, engine_config: {}, rubocop_config_store:)
+        @code = code
+        @exclude_paths = engine_config["exclude_paths"] || []
+        @include_paths = engine_config["include_paths"]
+        @rubocop_config_store = rubocop_config_store
+      end
+
+      def expanded_list
+        if @include_paths
+          include_based_files_to_inspect
+        else
+          exclude_based_files_to_inspect
+        end
+      end
+
+      protected
+
+      def exclude_based_files_to_inspect
+        rubocop_runner.send(:find_target_files, []).reject do |path|
+          exclude_due_to_config?(path)
+        end
+      end
+
+      def exclude_due_to_config?(path)
+        @exclude_paths.include?(local_path(path))
+      end
+
+      def include_based_files_to_inspect
+        @include_paths.map do |path|
+          if path =~ %r{/$}
+            rubocop_runner.send(:find_target_files, [path])
+          elsif rubocop_file_to_include?(path)
+            path
+          end
+        end.flatten.compact
+      end
+
+      def local_path(path)
+        realpath = Pathname.new(@code).realpath.to_s
+        path.gsub(%r|^#{realpath}/|, '')
+      end
+
+      def rubocop_file_to_include?(file)
+        if file =~ /\.rb$/
+          true
+        else
+          dir = File.dirname(file)
+          basename = File.basename(file)
+          @rubocop_config_store.for(dir).file_to_include?(basename)
+        end
+      end
+
+      def rubocop_runner
+        @rubocop_runner ||= RuboCop::Runner.new({}, @rubocop_config_store)
+      end
+    end
+  end
+end
--- a/lib/cc/engine/file_list_resolver.rb
+++ b/lib/cc/engine/file_list_resolver.rb
@@ -16,7 +16,7 @@ module CC
         end
       end
 
-      protected
+      private
 
       def exclude_based_files_to_inspect
         rubocop_runner.send(:find_target_files, []).reject do |path|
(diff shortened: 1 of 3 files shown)


lib/cc/engine/file_list_resolver.rb
```diff
@@ -0,0 +1,61 @@
...
+
+      def expanded_list
+        if @include_paths
+          include_based_files_to_inspect
+        else
+          exclude_based_files_to_inspect
+        end
+      end
+
+      protected
```
pbrisbin: I think you can go `private` here.

lib/cc/engine/rubocop.rb
```diff
@@ -26,25 +27,17 @@ def run
...
-        if @engine_config["include_paths"]
-          runner.send(:find_target_files, @engine_config["include_paths"])
-        else
-          runner.send(:find_target_files, []).reject do |path|
-            exclude_due_to_config?(path)
-          end
-        end
+        @file_list_resolver = FileListResolver.new(
+          code: @code, engine_config: @engine_config,
+          rubocop_config_store: rubocop_config_store
```
pbrisbin: :lipstick: each argument on its own line

lib/cc/engine/rubocop.rb
```diff
@@ -60,11 +53,13 @@ def local_path(path)
 
       def rubocop_config_store
         @rubocop_config_store ||= begin
-          config_store = RuboCop::ConfigStore.new
-          if (config_file = @engine_config["config"])
-            config_store.options_config = config_file
+          Dir.chdir(@code) do
+            config_store = RuboCop::ConfigStore.new
+            if (config_file = @engine_config["config"])
```
pbrisbin: I'm a little concerned that this config handling is wrong. I don't know enough about the SPEC and implementation thereof to say how it's supposed to work but I do know this:

There is code in codeclimate-yaml that turns `config: "some-file"` into `config: { file: "some-file" }` because want the config to always be an object:

```
irb(main):007:0> config = CC::Yaml.parse(<<EOYAML)
irb(main):008:1" engines:
irb(main):009:1"   rubocop:
irb(main):010:1"     enabled: true
irb(main):011:1"     config: ".cc-rubocop.yml"
irb(main):012:1" EOYAML
=> #<CC::Yaml::Nodes::Root:{"engines"=>{"rubocop"=>{"enabled"=>true}}}>
irb(main):013:0> config.engines["rubocop"].config
=> {"file"=>".cc-rubocop.yml"}
```

I think that means you want `@engine_config["config"]["file"]` here.

**BUT**, re-reading the spec, it sounds like the `config` object defined for the engine will be injected with the "default" keys, (`include_paths`, `exclude_paths`) and written as a single object in `/config.json`. So _I think_ that means you'd really want `@config["file"]` here, yeah?

@mrb can you clarify here, and then maybe clarify in the SPEC? We should probably get audit all the engines configuration handling too to be sure it's all working the way we intended.

@fhwang I'm bringing this up now because I'm noticing here, but since you're not changing this aspect of things there's no need to hold up the PR -- as long as we get something in Trello to come back to it.

lib/cc/engine/file_list_resolver.rb
```diff
@@ -0,0 +1,61 @@
...
+        realpath = Pathname.new(@code).realpath.to_s
+        path.gsub(%r|^#{realpath}/|, '')
+      end
+
+      def rubocop_file_to_include?(file)
+        if file =~ /\.rb$/
+          true
+        else
+          dir = File.dirname(file)
+          basename = File.basename(file)
```
pbrisbin: nit: how about `dir, basename = File.split(file)`?

spec/cc/engine/rubocop_spec.rb
```diff
@@ -239,6 +239,29 @@ def method
         assert !includes_check?(output, "Style/AndOr")
       end
 
+      it "ignores non-Ruby files even when passed in as include_paths" do
+        config_yml = "foo:\n  bar: \"baz\""
+        create_source_file("config.yml", config_yml)
+        output = run_engine(
+          "include_paths" => %w(config.yml)
+        )
+        assert !issues(output).detect { |i|
```
pbrisbin: :lipstick: `assert !` => `refute`

pbrisbin: :lipstick: `do`/`end` for multi-line blocks

spec/cc/engine/rubocop_spec.rb
```diff
@@ -239,6 +239,29 @@ def method
         assert !includes_check?(output, "Style/AndOr")
       end
 
+      it "ignores non-Ruby files even when passed in as include_paths" do
+        config_yml = "foo:\n  bar: \"baz\""
+        create_source_file("config.yml", config_yml)
+        output = run_engine(
+          "include_paths" => %w(config.yml)
```
pbrisbin: :lipstick: we prefer `%w[` for word arrays (note: you may have to tweak rubocop.yml like [this](https://github.com/codeclimate/builder/blob/master/.rubocop.yml#L44))

pbrisbin:
Bunch of style stuff and a big (but non-blocking) question. LGTM besides.

lib/cc/engine/rubocop.rb
mrb: Okay. I read this enough times and I understand it. I think we need to figure out what we want before we can decide how to proceed. I'll round up with Francis.

PR #21, opened 2015-09-21
Add RuboCop configuration and basic style corrections

andyw8:
It seems only fitting that the RuboCop engine should pass its own checks!

I added a minimal configuration based on the current code, then ran
RuboCop using the `-a` (autocorrect) option.

The only non-stylistic change I made was to use an `any?` block in place
of a `detect` block combined with `!!`.

There are two violations remaining, both about method length, but
since there’s a little more involved I’ve left them alone.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -2,7 +2,7 @@ require 'rake/testtask'
 
 Rake::TestTask.new do |t|
   t.test_files = Dir.glob('spec/**/*_spec.rb')
-  t.libs = ["lib", "spec"]
+  t.libs = %w[lib spec]
 end
 
 task(default: :test)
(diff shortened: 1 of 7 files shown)


Rakefile
```diff
@@ -2,7 +2,7 @@ require 'rake/testtask'
 
 Rake::TestTask.new do |t|
   t.test_files = Dir.glob('spec/**/*_spec.rb')
-  t.libs = ["lib", "spec"]
+  t.libs = %w(lib spec)
```
pbrisbin: We actually prefer `%w[ ]` internally, can you configure rubocop to prefer that and use it instead?

```yaml
Style/PercentLiteralDelimiters:
  PreferredDelimiters:
    '%w': []
    '%W': []
```

pbrisbin:
This all LGTM (modulo the `%w[]` change). Thanks for doing this!

andyw8:
Updated without your suggestions, @pbrisbin.

wfleming:
Thanks @andyw8!

PR #22, opened 2015-09-23
Tune remediation points

grstk:


lib/cc/engine/rubocop.rb
```diff
@@ -117,6 +118,27 @@ def content_body(cop_name)
         )
         File.exist?(path) ? File.read(path) : nil
       end
+
+      def remediation_points(violation)
+        formula = points_hash[violation.cop_name]
```
andyw8: Should there be a default value for when the rule is not found in the YAML?

jpignata:
Closing due to staleness.

PR #23, opened 2015-09-26
Extract `ViolationHash` class

andyw8:
I've been studying the RuboCop engine as I work on an engine for Reek.

There's a lot happening in `rubocop.rb`, with many mixed levels of
abstraction. It seems that extracting out the responsibility of
building the violation hash would make things easier to understand and
maintain.

I've not yet updated the tests to reflect the new structure. I'll await
feedback on this proposal first.

### DIFF
--- a/lib/cc/engine/violation_hash.rb
+++ b/lib/cc/engine/violation_hash.rb
@@ -1,0 +1,72 @@
+require "rubocop"
+require "active_support"
+require "active_support/core_ext"
+require "cc/engine/category_parser"
+
+module CC
+  module Engine
+    class ViolationHash
+      def initialize(violation, local_path)
+        @violation = violation
+        @local_path = local_path
+      end
+
+      def to_json
+        result = {
+          type: "Issue",
+          check_name: "Rubocop/#{violation.cop_name}",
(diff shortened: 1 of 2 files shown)


lib/cc/engine/violation_hash.rb
```diff
@@ -0,0 +1,72 @@
...
+          end: {
+            column: last_column + 1,
+            line: last_line,
+          }
+        }
+      end
+
+      def content_body(cop_name)
+        path = File.expand_path(
+          "../../../../config/contents/#{cop_name.underscore}.md", __FILE__
```
BlakeWilliams: Thoughts on making `"../../../../config/contents/` a constant at the top of the file?

andyw8: I think it should be extracted, but maybe in a followup PR? I'm a little wary of making changes at the same time as moving this fairly large chunk of code around since those can't be easily shown in a diff.

BlakeWilliams: That's reasonable. :+1:

jpignata:
Closing due to staleness.

PR #26, opened 2015-09-30
Add rake task to scrape rubocop documentation

BlakeWilliams:
This adds the `docs:scrape` rake task which will pull down the rubocop
repo and scrape all of the cops for class documentation and insert it
into markdown files.

This uses regular expresses to extract and manipulate the class
documentation since working with YARD and RDoc like this isn't
straightforward.


GordonDiggs:
Can you make this rake task a part of the Dockerfile build process?

mrb:
@GordonDiggs @BlakeWilliams We should make a way to have the docs task pinned to a specific repo/SHA.

BlakeWilliams:
Would something like this work?

```
TAG = 0.34.2
%x{cd rubocop-git && git checkout tags/#{TAG}}
```

BlakeWilliams:
@GordonDiggs Would adding a `RUN` command to run this task in the `Dockerfile` be an acceptable way of handling that?

mrb:
@BlakeWilliams Looks reasonable to me yes

GordonDiggs:
@BlakeWilliams yep exactly. Make sure it's after the point at which we copy the repo's files into the image.

And, now that it's part of the build process, we can probably remove the `md` files from the repo itself.

BlakeWilliams:
@GordonDiggs Great idea, added.

Dockerfile
```diff
@@ -8,9 +8,16 @@ RUN apk --update add ruby ruby-dev ruby-bundler build-base && \
     bundle install -j 4 && \
     apk del build-base && rm -fr /usr/share/ri
 
+RUN apk update
```
GordonDiggs: Can these 2 be rolled into the `RUN` command above?

BlakeWilliams:
Tests fail locally now unless you run `rake scrape:docs` before the tests. Thoughts on migrating the tests to run in the docker container?

BlakeWilliams:
@GordonDiggs Can I get a re-review or +1 when you get a chance?

Dockerfile
```diff
@@ -6,11 +6,16 @@ COPY Gemfile.lock /usr/src/app/
 
 RUN apk --update add ruby ruby-dev ruby-bundler build-base && \
     bundle install -j 4 && \
-    apk del build-base && rm -fr /usr/share/ri
+    apk del build-base && rm -fr /usr/share/ri && \
+    apk --update add git
```
GordonDiggs: Is there a specific reason to do this here rather than adding `git` to the list above?

circle.yml
```diff
@@ -7,8 +7,8 @@ machine:
 
 test:
   override:
-    - bundle exec rake
     - docker build -t=$PRIVATE_REGISTRY/$CIRCLE_PROJECT_REPONAME:b$CIRCLE_BUILD_NUM .
+    - docker run $PRIVATE_REGISTRY/$CIRCLE_PROJECT_REPONAME:b$CIRCLE_BUILD_NUM bundle exec rake
```
GordonDiggs: cool

Dockerfile
BlakeWilliams: Just separating it from the ruby specific packages. I can add it in to the above if that's preferable.

GordonDiggs: Yeah we generally just group all the packages together into one `add` call

BlakeWilliams: Added and pushed.

GordonDiggs:
LGTM

PR #27, opened 2015-10-06
Conditionally include Rails cops

jpignata:
Only include the Rails-specific cops if we see railties in the Gemfile.lock file in the project. This should resolve cases where Rails-specific advice is given for projects that aren't using Rails (e.g. Rails/Delegate).

### DIFF
--- a/lib/cc/engine/rubocop.rb
+++ b/lib/cc/engine/rubocop.rb
@@ -4,6 +4,7 @@ require "rubocop"
 require "rubocop/cop/method_complexity_patch"
 require "cc/engine/category_parser"
 require "cc/engine/file_list_resolver"
+require "cc/engine/lockfile_specs"
 require "active_support"
 require "active_support/core_ext"
 
@@ -66,7 +67,7 @@ module CC
 
       def rubocop_team_for_path(path)
         rubocop_config = rubocop_config_store.for(path)
-        RuboCop::Cop::Team.new(RuboCop::Cop::Cop.all, rubocop_config)
+        RuboCop::Cop::Team.new(cops, rubocop_config)
       end
 
       def violation_positions(location)
@@ -116,6 +117,16 @@ module CC
           "../../../../config/contents/#{cop_name.underscore}.md", __FILE__
         )
         File.exist?(path) ? File.read(path) : nil
+      end
+
+      def cops
+        specs = LockfileSpecs.new("Gemfile.lock")
+
+        if specs.include?("railties")
+          RuboCop::Cop::Cop.all
+        else
+          RuboCop::Cop::Cop.non_rails
+        end
       end
     end
   end
--- a/lib/cc/engine/lockfile_specs.rb
+++ b/lib/cc/engine/lockfile_specs.rb
@@ -1,0 +1,18 @@
+require "bundler/lockfile_parser"
+
+class LockfileSpecs
+  def initialize(lockfile)
+    if File.exist?(lockfile)
+      @contents = File.read(lockfile)
+    else
+      @contents = ""
+    end
+  end
+
+  def include?(name)
+    @lockfile ||= Bundler::LockfileParser.new(@contents)
+    @lockfile.specs.any? { |spec| spec.name == name }
+  rescue Bundler::LockfileError
+    false
+  end
+end
--- a/spec/cc/engine/lockfile_specs_spec.rb
+++ b/spec/cc/engine/lockfile_specs_spec.rb
@@ -1,0 +1,37 @@
+require "spec_helper"
+require "cc/engine/lockfile_specs"
+
+module CC::Engine
+  describe LockfileSpecs do
+    describe "#include?" do
+      before do
+        @specs = LockfileSpecs.new("./spec/fixtures/Gemfile.lock")
+      end
+
+      it "returns true when spec is include in Gemfile.lock" do
+        @specs.include?("ansi").must_equal(true)
+        @specs.include?("i18n").must_equal(true)
+        @specs.include?("astrolabe").must_equal(true)
+      end
+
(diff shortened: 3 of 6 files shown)


GordonDiggs:
LGTM save for the CC issues, which I agree with

lib/cc/engine/rubocop.rb
```diff
@@ -117,6 +119,18 @@ def content_body(cop_name)
         )
         File.exist?(path) ? File.read(path) : nil
       end
+
+      def cops
+        if File.exists?("Gemfile.lock")
+          detector = GemDetector.new(File.read("Gemfile.lock"))
```
pbrisbin: WDYT about letting `GemDetector` check existence and read the file itself? Then this method could be a little more squint-friendly:

```rb
detector = GemDetector.new("Gemfile.lock")

if detector.gem?("railties")
  RuboCop::Cop::Cop.all
else
  RuboCop::Cop::Cop.non_rails
end
```

spec/cc/engine/gem_detector_spec.rb
```diff
@@ -0,0 +1,25 @@
+require "spec_helper"
+require "cc/engine/gem_detector"
+
+module CC::Engine
+  describe GemDetector do
+    describe "#has_gem?" do
+      before do
+        gemfile = File.read("Gemfile.lock")
```
pbrisbin: How about copying our current lockfile to a fixture so changes to our gems don't risk breaking this spec?

pbrisbin:
This is great. Some minor nits, but LGTM besides.

andyw8:
Might it be confusing if running RuboCop locally, where it doesn't auto-detect Rails, gives different results to running on Code Climate?

jpignata:
@andyw8 Yeah, maybe. I opened this because RuboCop locally does not include the Rails cops by default. You need to pass -R explicitly to get output from the Rails cops. It seemed like Code Climate was acting incorrectly by always including them. Does that potentially make it less confusing? Do you still see an issue?

spec/cc/engine/gem_detector_spec.rb
jpignata: :+1:

Done.

lib/cc/engine/rubocop.rb
jpignata: :+1:

Done.

jpignata:
@GordonDiggs :+1: Addressed RuboCop's feedback by running it locally. I can't seem to access the Code Climate feed for this project.

spec/cc/engine/gem_detector_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  describe GemDetector do
+    describe "#present?" do
+      before do
+        @detector = GemDetector.new("./spec/fixtures/Gemfile.lock")
+      end
+
+      it "returns true when spec is present in Gemfile.lock" do
+        @detector.present?("ansi").must_equal(true)
+        @detector.present?("i18n").must_equal(true)
+        @detector.present?("astrolabe").must_equal(true)
```
pbrisbin: I consider it a bit of a smell when you have a nice predicate method but then resort to checking equality with `true`/`false` in your specs. I think things would read better like this:

```rb
specs = LockfileSpecs.new("./spec/fixures/Gemfile.lock")

expect(specs).to include("ansi")
expect(specs).not_to include("pants")
```

Renaming something for the specs alone isn't always the best thing to do, but in this case I think it'd be more readable in the implementation code as well.

What do you think about renaming `GemDetector` => `LockfileSpecs` and `present?` => `include?`. I think that's all it'd take.

lib/cc/engine/gem_detector.rb
```diff
@@ -0,0 +1,17 @@
...
+
+class GemDetector
+  def initialize(lockfile)
+    @contents = File.read(lockfile)
+  end
+
+  def present?(name)
+    begin
+      @lockfile ||= Bundler::LockfileParser.new(@contents)
+    rescue Bundler::LockfileError
```
pbrisbin: I think it's safe to move this `rescue` to the end of the method so you can ditch the explicit `begin` and `return`

```rb
def present?(name)
  @lockfile ...
  @lockfile.any? ...
rescue ...
  false
end
```

lib/cc/engine/gem_detector.rb
```diff
@@ -0,0 +1,17 @@
+require "bundler/lockfile_parser"
+
+class GemDetector
+  def initialize(lockfile)
+    @contents = File.read(lockfile)
```
pbrisbin: WDYT about doing the `File.exist?` check here (and setting an empty `@content` if not present)?

*Slightly* dirty implicitly triggering the bundler exception=>false behavior with the empty content, but it would allow `Rubcop#cops` to be a cleaner one-level if-else, which I think is worth it.

spec/cc/engine/gem_detector_spec.rb
jpignata: I'm not against the rename, but I'm not sure that predicate-aware syntax is available as we're using minitest. There may be some kind of matcher gem that introduces it but I'd argue it isn't worth it.

lib/cc/engine/gem_detector.rb
jpignata: Into it.

andyw8:
@jpignata Yes, fixing the current behaviour certainly makes sense, I was just wondering about the overall  auto-detecting approach, but I suppose that's really a separate discussion.

spec/cc/engine/gem_detector_spec.rb
pbrisbin: *shakes fist at minidoc*

Agreed not worth it. Still in favor of there rename though.

pbrisbin:
This LGTM

PR #30, opened 2015-10-22
Add configurable remediation points from yaml

BlakeWilliams:
This adds `config/cops.yml` which contains the configuration of the
points that each cop should return.

This also ups the default remediation points value to `200,000` from `50,000` since the previous value was too small to affect ratings.

### DIFF
--- a/config/cops.yml
+++ b/config/cops.yml
@@ -1,0 +1,1 @@
+Metrics/ClassLength: 400_000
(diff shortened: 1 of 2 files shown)


config/cops.yml
```diff
@@ -0,0 +1,246 @@
+Style/AccessModifierIndentation: 200_000
```
GordonDiggs: Why list them if they're at the default?

BlakeWilliams: I just assumed that these would be tweaked in the near future so I added them in real quick.

If we want to only include cops that we change I can remove all of these for now.

GordonDiggs:
LGTM

PR #31, opened 2015-10-23
Use Regexp match to detect gem

jpignata:
It turns out coupling to Bundler::LockfileParser was a bad idea
and has caused odd behavior in practice. This change simplies
the approach and just looks through the lines of the raw
Gemfile.lock and looks for railties instead of using any
bundler components.

cc @wfleming

### DIFF
--- a/lib/cc/engine/lockfile_specs.rb
+++ b/lib/cc/engine/lockfile_specs.rb
@@ -1,15 +1,13 @@
-require "bundler"
-
 class LockfileSpecs
   def initialize(lockfile)
-    @contents = File.read(lockfile) if File.exist?(lockfile)
+    @lockfile = lockfile
   end
 
   def include?(name)
-    return false if @contents.nil?
-    @lockfile ||= Bundler::LockfileParser.new(@contents)
-    @lockfile.specs.any? { |spec| spec.name == name }
-  rescue Bundler::LockfileError
-    false
+    return false unless File.exist?(@lockfile)
+
+    File.open(@lockfile) do |file|
+      file.each_line.any? { |line| line =~ /#{name}/ }
+    end
   end
 end
--- a/spec/cc/engine/lockfile_specs_spec.rb
+++ b/spec/cc/engine/lockfile_specs_spec.rb
@@ -20,13 +20,6 @@ module CC::Engine
         @specs.include?("python").must_equal(false)
       end
 
-      it "returns false when a LockfileError occurs while parsing" do
-        fixture = "./spec/fixtures/Gemfile.lock.with_merge_conflicts"
-        specs = LockfileSpecs.new(fixture)
-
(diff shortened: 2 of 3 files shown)


lib/cc/engine/lockfile_specs.rb
```diff
@@ -1,15 +1,14 @@
...
-    return false if @contents.nil?
-    @lockfile ||= Bundler::LockfileParser.new(@contents)
-    @lockfile.specs.any? { |spec| spec.name == name }
-  rescue Bundler::LockfileError
-    false
+    if File.exist?(@lockfile)
+      @contents ||= File.read(@lockfile)
+      @contents.lines.any? { |line| line =~ /#{name}/ }
+    else
+      false
```
pbrisbin: I think this is redundant.

lib/cc/engine/lockfile_specs.rb
```diff
@@ -1,15 +1,14 @@
...
 
   def include?(name)
-    return false if @contents.nil?
-    @lockfile ||= Bundler::LockfileParser.new(@contents)
-    @lockfile.specs.any? { |spec| spec.name == name }
-  rescue Bundler::LockfileError
-    false
+    if File.exist?(@lockfile)
+      @contents ||= File.read(@lockfile)
+      @contents.lines.any? { |line| line =~ /#{name}/ }
```
pbrisbin: I think if you use `File.open` you can avoid loading the whole file into memory.

```rb
File.open(@lockfile) { |f| f.lines.any? { ... } }
```

If that matters

jpignata: Probably better. Changed it.

spec/cc/engine/lockfile_specs_spec.rb
```diff
@@ -20,16 +20,8 @@ module CC::Engine
...
-      it "returns false when a LockfileError occurs while parsing" do
-        fixture = "./spec/fixtures/Gemfile.lock.with_merge_conflicts"
-        specs = LockfileSpecs.new(fixture)
-
-        specs.include?("ansi").must_equal(false)
-      end
-
       it "returns false when no Gemfile.lock file is found" do
         specs = LockfileSpecs.new(SecureRandom.hex(32))
-
```
pbrisbin: :lipstick: I like this line. Our styleguide says to separate the 4 phases of specs with whitespace.

pbrisbin:
Minor :lipstick: but LGTM

spec/cc/engine/lockfile_specs_spec.rb
jpignata: :+1:

jpignata:
Actually, looking into this more today, it seems like this bit of code isn't even necessary since the default configuration generated by `codeclimate init` provides really sensible defaults. I'll follow up with a clean-up pull.

PR #32, opened 2015-10-26
Allow remediation points to be multipliable

BlakeWilliams:
This adds a check for the format of `[score/limit]` in a cop message and
will extract the `score` value and multiply it by the defined remediation
points.

Remediation points for metric cops were also tweaked to make up for the addition of multiplication.

### DIFF
--- a/config/cops.yml
+++ b/config/cops.yml
@@ -1,1 +1,24 @@
-Metrics/ClassLength: 400_000
+Metrics/AbcSize:
+  violation_points: 100_000
+Metrics/BlockNesting:
+  base_points: 30_000
+Metrics/ClassLength:
+  base_points: 400_000
+Metrics/CyclomaticComplexity:
+  base_points: 75_000
+  violation_points: 10_000
+Metrics/LineLength:
+  base_points: 20_000
+  violation_points: 5_000
+Metrics/MethodLength:
+  base_points: 50_000
+  violation_points: 10_000
+Metrics/ModuleLength:
+  base_points: 500_000
+  violation_points: 5_000
+Metrics/ParameterList:
+  base_points: 10_000
+  violation_points: 50_000
+Metrics/PerceivedComplexity:
+  base_points: 50_000
+  violation_points: 5000
--- a/lib/cc/engine/violation_decorator.rb
+++ b/lib/cc/engine/violation_decorator.rb
@@ -1,0 +1,56 @@
+require 'safe_yaml'
+
+module CC
+  module Engine
+    class ViolationDecorator < SimpleDelegator
+      MULTIPLIER_REGEX = %r{\[([\d\.]+)\/([\d\.]+)\]}
+
+      DEFAULT_REMEDIATION_POINTS = 200_000
+      DEFAULT_POINTS_PER_VIOLATION = 50_000
+
+      def remediation_points
+        if multiplier?
+          base_points + violation_points
+        else
+          base_points
+        end
+      end
+
+      private
+
+      def base_points
+        cop_list.
+          fetch(cop_name, {}).
+          fetch("base_points", DEFAULT_REMEDIATION_POINTS)
+      end
+
+      def violation_points
+        per_violation_points = cop_list.
+          fetch(cop_name, {}).
+          fetch("violation_points", DEFAULT_POINTS_PER_VIOLATION)
+
+        per_violation_points * multiplier
+      end
+
+      def multiplier
+        result = message.scan(MULTIPLIER_REGEX)
+        score, max = result[0]
+        score.to_i - max.to_i
+      end
+
+      def multiplier?
+        message.match(MULTIPLIER_REGEX)
+      end
+
+      def cop_list
+        return @cop_list if @cop_list
+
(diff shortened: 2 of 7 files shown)


lib/cc/engine/violation_decorator.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      def cop_list
+        return @cop_list if @cop_list
+
+        cops_path = File.expand_path(
+          File.join(File.dirname(__FILE__), "../../../config/cops.yml")
+        )
+
+        File.open(cops_path, "r") do |file|
+          @cop_list = YAML.load(file.read)
```
pbrisbin: Can we use `SafeYAML` here? Even though it's not user input, better to just always use that everywhere I think. 

Also, I think there's a `.load_file` method (on both modules).

lib/cc/engine/violation_decorator.rb
```diff
@@ -0,0 +1,41 @@
...
+        if multiplier?
+          points * multiply_by
+        else
+          points
+        end
+      end
+
+      private
+
+      def multiply_by
```
GordonDiggs: WDYT of calling this method `multiplier`? Mathematically, that's its function

GordonDiggs: I think the `multipler?` and `multiplier` pattern works well too

lib/cc/engine/violation_decorator.rb
```diff
@@ -0,0 +1,41 @@
...
+        result = message.scan(MULTIPLIER_REGEX)
+        result[0][0].to_i
```
GordonDiggs: Is there a way to get these from the object itself? I assume they have to exist as attributes somewhere for the message to be generated.

lib/cc/engine/violation_decorator.rb
BlakeWilliams: Agreed.

lib/cc/engine/violation_decorator.rb
BlakeWilliams: I'll move it over.

pbrisbin:
> This adds a check for the format of [score/limit] in a cop message and
will extract the score value and multiply it by the defined remediation
points.

I'm not sure this logic makes sense.

- I'd image scaling by *overage*, not raw *score* is better (`110/100` shouldn't hurt you that much more than `11/10` right?)
- Even then, simple multiplication is probable not right (`11/10` won't impact points in such a case)

I think there is a scaling scheme in legacy analyzer code somewhere that would be closer to what we want:

https://github.com/codeclimate/analyzer/blob/9f96fa2f8c65cc262bedc4a95c5d36091dc7db28/lib/quality/threshold_penalty.rb

This describes things in terms of base-cost, cost-per, and overage. I think base-cost and cost-per are cop-specific and overage could be parsed out of the message.

BlakeWilliams:
That makes sense, I originally had it to only calculate overage but that didn't feel great.

I'll try this and and tweak things until they feel reasonable.

lib/cc/engine/violation_decorator.rb
BlakeWilliams: It just returns an array. It's super misleading and the alternative is to use named capture groups and `=~` which defines the capture group names as locals.

Not sure I have an opinion on which is better.

GordonDiggs: Sorry - I wasn't clear in my comment. I mean is there a way to get the score and threshold from the violation object?

BlakeWilliams: Ah, gotcha.

Unfortunately it seems like it only comes through `message`. https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/offense.rb

BlakeWilliams:
@pbrisbin Updated to act more like the code you linked. The results I got from running it feel largely the same as before, which I assume is a good thing.

pbrisbin:
@BlakeWilliams did you forget to push again?

BlakeWilliams:
Yep, alias getting in the way. Removing it because it's clearly a recurring problem.

config/cops.yml
```diff
@@ -1 +1,24 @@
-Metrics/ClassLength: 400_000
+Metrics/AbcSize:
+  violation_cost: 100_000
```
pbrisbin: This is my fault for not mentioning, but these are `_points` not `_cost` these days :)

BlakeWilliams: I was somewhat using them interchangeably. Do we prefer to rename these to points?

pbrisbin:
Other than the old "cost" naming, this LGTM.

You can update the `.rubocopy.yml` to enforce the trailing-dot style instead. It's what we prefer anyway.

BlakeWilliams:
I prefer it as well. Thanks for the review!

config/cops.yml
pbrisbin: Definitely.

Both are terms and they mean different things. They're related as something like `cost = points / 100_000`.

Clearly, maintaining two values when one can be inferred from the other is kind of dumb, and we're on a pledge to only use "points" from now on.

PR #34, opened 2015-10-30
Always include all Cops by default

jpignata:
I made a change a couple of weeks back to only include Rails cops in
the event we detect railties in the project's Gemfile.lock. I'm
convinced this was the wrong course of action since our default rubocop
[configuration file](https://github.com/codeclimate/codeclimate/blob/master/config/rubocop/.rubocop.yml#L1119) has these mostly disabled out of the box. It makes much
more sense to rely on either a user's existing .rubocop.yml or our default
rather than mucking with this within the engine itself.


wfleming:
:+1: on this direction. Seems like more of this should get delegated to the "configure" phase (though as that phase needs to be smarter it seems like we'll eventually want to just ask particular engines how they should be configured rather than baking it into our gem).

PR #35, opened 2015-11-03
Run docker build as a dependency, not a test step

pbrisbin:
Not overriding the dependency step can cause Circle to infer ruby and run a
bundle install. That can leave files in the source tree that end up copied into
the docker image and mess with bundler's operation within that image.

Also, the build step is (IMO) a dependency, not itself a test.

/cc @codeclimate/review -- open to opinions here


ABaldwinHunter:
LGTM

BlakeWilliams:
:shipit:

PR #36, opened 2015-11-06
SafeYAML: set safe mode by default

wfleming:
we're loading SafeYAML, but then don't actually specify safe mode when
loading. We should just set it to the default when we load it.

cc @codeclimate/review


fhwang:
LGTM

PR #38, opened 2015-11-10
Monkey patch class and method length cops

BlakeWilliams:
Rubocop `ClassLength` and `MethodLength` cops have a function called
`check_code_length` that determines the lines to report for the
violation.

Rubocop passes in `keyword` as the method to be called on `node`
which always returns `1` since it's grabbing the defining keyword, eg:
`class` or `def`.

This monkey patch passes in the node's `loc` property directly to return
the correct lines.


BlakeWilliams:
cc @codeclimate/review

fhwang:
I'm not super-familiar with Rubocop internals, but: Would this make more sense as a PR to Rubocop itself? In the interests of avoiding custom logic that would it harder for us to keep up-to-date with future Rubocop changes.

wfleming:
[We used `module_eval` for the existing monkey patch](https://github.com/codeclimate/codeclimate-rubocop/blob/master/lib/rubocop/cop/method_complexity_patch.rb#L2) so that if rubocop internals change in a future version, we would find out very quickly: I think we should do the same here.

I agree with @fhwang that we should submit patches to rubocop where realistic (we recently did submit a patch for a bug). If this is adding functionality that is specific to our usage of rubocop, though, we should probably kick that can down the road a bit. Before we start submitting patches to them that are specifically for our use case, we should probably get @mrb involved in first having a discussion with their team around closer collaboration/rubocop behaving more like a "native engine", etc.

BlakeWilliams:
I'll move it over to `module_eval`.

I'm not sure if it's a bug or not but I do agree it's better to make a patch on their end rather than ours. Thoughts on migrating this to `module_eval` for now and I'll follow up with a Rubocop PR?

BlakeWilliams:
Talked about this in Slack. Making a PR to Rubocop and merging for now.

PR #42, opened 2015-12-22
Tune remediation points

ABaldwinHunter:
Tune complexity points, following calculation function in [codeclimate-eslint](https://github.com/codeclimate/codeclimate-eslint/blob/master/lib/checks.js#L117), creating more parity with Ruby analysis in Code Climate classic.

cc @codeclimate/review @noahd1

### DIFF
--- a/config/cops.yml
+++ b/config/cops.yml
@@ -1,24 +1,25 @@
 Metrics/AbcSize:
-  violation_points: 100_000
+  base_points: 1_000_000
+  violation_points: 70_000
 Metrics/BlockNesting:
-  base_points: 30_000
+  base_points: 100_000
 Metrics/ClassLength:
-  base_points: 400_000
-Metrics/CyclomaticComplexity:
-  base_points: 75_000
-  violation_points: 10_000
+  base_points: 5_000_000
+  violation_points: 35_000
+Metrics/CyclomaticComplexity: # This check is per method
+  base_points: 1_000_000
+  violation_points: 70_000
 Metrics/LineLength:
-  base_points: 20_000
-  violation_points: 5_000
+  base_points: 50_000
 Metrics/MethodLength:
-  base_points: 50_000
-  violation_points: 10_000
+  base_points: 1_000_000
+  violation_points: 70_000
 Metrics/ModuleLength:
-  base_points: 500_000
-  violation_points: 5_000
+  base_points: 5_000_000
+  violation_points: 35_000
 Metrics/ParameterList:
-  base_points: 10_000
-  violation_points: 50_000
+  base_points: 500_000
+  violation_points: 100_000
 Metrics/PerceivedComplexity:
-  base_points: 50_000
-  violation_points: 5000
+  base_points: 1_000_000
+  violation_points: 70_000
--- a/spec/cc/engine/rubocop_spec.rb
+++ b/spec/cc/engine/rubocop_spec.rb
@@ -323,7 +323,7 @@ module CC::Engine
       def includes_content_for?(output, cop_name)
         issue = issues(output).detect { |i| i["check_name"] =~ /#{cop_name}$/ }
 
-        issue["content"]["body"].present?
+        issue["content"] && issue["content"]["body"].present?
       end
 
       def issues(output)


spec/cc/engine/rubocop_spec.rb
```diff
@@ -323,7 +323,7 @@ def includes_check?(output, cop_name)
       def includes_content_for?(output, cop_name)
         issue = issues(output).detect { |i| i["check_name"] =~ /#{cop_name}$/ }
 
-        issue["content"]["body"].present?
+        issue["content"] && issue["content"]["body"].present?
```
ABaldwinHunter: This change isn't related, but came up when running tests. Happy to remove for cleaner diff if preferred.

pbrisbin: I'm fine with this drive-by, but I'd make it its own commit if you can.

config/cops.yml
```diff
@@ -1,24 +1,25 @@
...
-  base_points: 30_000
+  base_points: 100_000
 Metrics/ClassLength:
-  base_points: 400_000
+  base_points: 800_000
+  violation_points: 75_000
 Metrics/CyclomaticComplexity:
-  base_points: 75_000
-  violation_points: 10_000
+  base_points: 1_000_000
```
mrb: Does this effectively mean that any instance of Cyclomatic complexity makes a file a "B"?

spec/cc/engine/rubocop_spec.rb
ABaldwinHunter: Noted. Moved to separate commit.

config/cops.yml
ABaldwinHunter: 2,000,000 remediation points are needed in penalties to go from A to B. 

So not always, but a lot of the time an instance of Cyclomatic Complexity could have that large of grade impact.

I was talking with @brynary yesterday and had the impression that that level of grade impact would be _close_ to what we want for complexity - but maybe requires some additional discussion?

ABaldwinHunter:
@codeclimate/review @brynary 

Ready for another look!

ABaldwinHunter:
from commit message:
```
In Classic Analysis, typical remediation point formular for complexity were
roughly:

Class
  5_000_000 + 50_000n

Method
  1_000_000 + 100_000n

This change updates codeclimate-rubocop to be closer to those standards.

Code Climate users can turn on one or all complexity checks, with the
understanding that there may be overlap between them.

Users can also configure thresholds in their .rubocop.yml file.
```

ABaldwinHunter:
Ready for re-review.

ABaldwinHunter:
cc @codeclimate/review

spec/cc/engine/rubocop_spec.rb
GordonDiggs: I still only see one commit on this branch

ABaldwinHunter: Ahhh sorry. 
Fixed. Then made additional changed and squashed everything again. My bad. Updating.

ABaldwinHunter: Done! Thanks for the catch.

GordonDiggs:
LGTM

config/cops.yml
```diff
@@ -1,24 +1,26 @@
...
+  violation_points: 35_000
+Metrics/CyclomaticComplexity: # This check is per method
+  base_points: 1_000_000
+  violation_points: 70_000
 Metrics/LineLength:
-  base_points: 20_000
-  violation_points: 5_000
-Metrics/MethodLength:
-  base_points: 50_000
+  base_points: 100_000
```
brynary: Is `LineLength` for a single LOC being too long (wide)? If so, this seems too high.

brynary:
Reviewed again. Had one question about `LineLength` (added inline). Everything else LGTM

config/cops.yml
ABaldwinHunter: Yes, concur that's the offense.

Reverting to the [current setting](https://github.com/codeclimate/codeclimate-rubocop/blob/master/config/cops.yml#L10):
```
Metrics/LineLength:
  base_points: 20_000
  violation_points: 5_000
```

if that sounds good?

brynary: So for a style thing (like missing semicolon in JS) we do 50k. I think that applies here as well. I'm not sure anything would ever be less than 50k, since the 50k number is intended to account for the time it takes to open the file, make the change, save, commit, push... it's about 5 mins even for something small.

For a long line, it doesn't seem to get harder to fix as the line gets longer. Therefore, I'd just do a flat 50k.

PR #43, opened 2015-12-29
Upgrade rubocop to fix Style/WordArray warning.

GordonDiggs:
A few users have run into this problem, and upgrading rubocop
(specifically the parsing dependencies) fixes the issue. Example engine
output is shown below.

Before this change
==================

```
An error occurred while Style/WordArray cop was inspecting
config/initializers/logs.rb.
To see the complete backtrace run rubocop -d.
```

After this change
=================

```
warning: parser/current is loading parser/ruby22, which recognizes
warning: 2.2.3-compliant syntax, but you are running 2.2.2.
warning: please see
https://github.com/whitequark/parser#compatibility-with-ruby-mri.
```

Note: this version warning is new, but as there were no syntax changes
introduced in ruby 2.2.3, I don't believe it is an issue.
(https://www.ruby-lang.org/en/news/2015/08/18/ruby-2-2-3-released/)

@codeclimate/review


Gemfile.lock
```diff
@@ -61,3 +63,6 @@ DEPENDENCIES
   rubocop
   rubocop-rspec
   safe_yaml
+
+BUNDLED WITH
+   1.10.6
```
ABaldwinHunter: Is the `Bundled with` addition intentional or impactful?

GordonDiggs:
@ABaldwinHunter does this interfere with your points work?

Gemfile.lock
GordonDiggs: Neither. Will remove though

ABaldwinHunter:
@GordonDiggs I don't think it interferes.

Minor comment but LGTM

ABaldwinHunter:
Also, the points work is already merged

ABaldwinHunter:
Thanks for heads up

PR #49, opened 2016-02-16
Update rubocop-rspec to 1.4.0

jpignata:
There are some incompatibilites between the latest version of RuboCop
and rubocop-rspec which were fixed in this release.

See: https://github.com/nevir/rubocop-rspec/issues/62

@codeclimate/review


wfleming:
Is it feasible to add a spec for whatever was broken?

Other than that :shipit:.

Issue #39, opened 2015-11-20
Configure custom rubocop.yml file


ivdma:
couldn't figure out how to set a path to a custom rubocop.yml file. Is it undocumented or is it broken? This is my config:

```
---
engines:
  rubocop:
    enabled: true
    config: '.custom_rubocop.yml'
```

driv3r:
we have our own `.rubocop.yml` file within the project, but doesn't quite work.

Also just an idea, it would probably make sense to merge `.rubocop.yml` file from project (i.e. with custom rules, but not all of them) with `.rubocop.yml` file from engine for a fallback. What do you think?

wfleming:
First, I apologize to both of you for how long it's taken us to get back to you on these issues.

@ivdma We do support an alternative rubocop config file, and your config looks correct to me (and a similar one worked for me locally). Is this still not working for you? If not, are you using the CLI or locally or running analysis on codeclimate.com? If you're using the CLI, the first thing I would try is running `codeclimate engines:install`: you may still have an older version of the engine from before this functionality was added.

@driv3r What about your own `.rubocop.yml` doesn't seem to be quite working?

Right now, we don't support importing one config file into another: normally when people want to do this, they're importing from something outside of the checked-in repo, which doesn't usually work because of how we isolate running engines. I'm hesitant to automatically merge Code Climate's default Rubocop rules into a checked-in `.rubocop.yml` since if we tweaked our default rules (which we do sometimes), that would change analysis results, which would be surprising & undesirable for many people who have customized their own `.rubocop.yml`.

We do hope to eventually support importing base config files via Rubocop's `inherit_from` config directive, and I think importing our default rules explicitly that way would be the way to do it, but we haven't figured out a good way to do it yet that doesn't break aspects of the limits we place on engines for security reasons.

driv3r:
@wfleming 

- I had only `.rubocop.yml`  file with out custom cops, but it didn't work (both on CC and locally with `codeclimate analyze`
- I've tried adding it as config option as @ivdma but with `codeclimate analyze` it didn't worked as well
- I've also tried changing name of the config file, without result

so currently I don't know any way of changing this config

wfleming:
@driv3r Would you mind sharing your `.rubocop.yml` & `.codeclimate.yml`? Feel free to email me at will@codeclimate.com if you'd rather not share those publicly. Thanks.

dfaulken:
[...]

PR #51, opened 2016-02-16
Resolve all paths to absolute paths

wfleming:
When paths were sent in as relative paths from the CLI, `Exclude`
directives within a rubocop config yml were not correctly applied. This
is because Rubocop internally turns those excludes into absolute paths,
and it seems like it was then trying to match the being-analyzed
relative path against the had-been-made-absolute excluded path.

@codeclimate/review haven't been able to get a spec breaking this, unfortunately. But have reproduced behavior & confirmed this fixes it. Thoughts?

### DIFF
--- a/lib/cc/engine/file_list_resolver.rb
+++ b/lib/cc/engine/file_list_resolver.rb
@@ -29,18 +29,22 @@ module CC
       end
 
       def include_based_files_to_inspect
-        @include_paths.map do |path|
-          if path =~ %r{/$}
+        absolute_include_paths.flat_map do |path|
+          if Dir.exist?(path)
             rubocop_runner.send(:find_target_files, [path])
           elsif rubocop_file_to_include?(path)
             path
           end
-        end.flatten.compact
+        end.compact
       end
 
       def local_path(path)
         realpath = Pathname.new(@root).realpath.to_s
         path.gsub(%r{^#{realpath}/}, '')
+      end
+
+      def absolute_include_paths
+        @include_paths.map { |path| Pathname.new(path).realpath.to_s }
       end
 
       def rubocop_file_to_include?(file)
--- a/spec/cc/engine/rubocop_spec.rb
+++ b/spec/cc/engine/rubocop_spec.rb
@@ -62,6 +62,37 @@ module CC::Engine
         expect(includes_check?(output, "Lint/UselessAssignment")).to be false
       end
 
+      it "respects excludes in an inherit_from directive" do
+        create_source_file("foo.rb", <<-EORUBY)
+          def method
+            unused = "x"
+            return false
+          end
+        EORUBY
+        create_source_file("bar.rb", <<-EORUBY)
+          def method
+            unused = 42
+            return true
+          end
+        EORUBY
+
+        create_source_file(
+          ".rubocop.yml",
+          "inherit_from: .rubocop_todo.yml\nAllCops:\n  DisabledByDefault: true\nLint/UselessAssignment:\n  Enabled: true\n"
+        )
+        create_source_file(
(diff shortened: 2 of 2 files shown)


lib/cc/engine/file_list_resolver.rb
```diff
@@ -29,13 +29,13 @@ def exclude_due_to_config?(path)
       end
 
       def include_based_files_to_inspect
-        @include_paths.map do |path|
-          if path =~ %r{/$}
+        @include_paths.map(&method(:abs_path)).flat_map do |path|
```
GordonDiggs: WDYT of

```ruby
@include_paths.flat_map do |pathname|
  path = abs_path(pathname)
  ....
end
```

jpignata: I'd be tempted to extract this as its own method and stop referencing the ivar.

dblandin:
> haven't been able to get a spec breaking this, unfortunately. But have reproduced behavior & confirmed this fixes it. Thoughts?

For posterity's sake, could you include the steps you took to reproduce and maybe the relevant snippet from the `.rubocop.yml` config file?

jpignata:
I don't suppose setting up a `./code/` directory and using `Dir.chdir` leads to a failing spec? I'm guessing you've been down those roads.

LGTM in any case.

lib/cc/engine/file_list_resolver.rb
wfleming: @jpignata not totally clear what you mean. Were you thinking something like `@include_paths.flat_map(&method(:expanded_include_path)).compact` as a final form?

jpignata: I meant adding a method to hide both `@include_paths` and what you're doing above with the method proc.

wfleming:
> I don't suppose setting up a ./code/ directory and using Dir.chdir leads to a failing spec? I'm guessing you've been down those roads.

@jpignata Yes, I have been. No dice, so far.

> For posterity's sake, could you include the steps you took to reproduce and maybe the relevant snippet from the .rubocop.yml config file?

@dblandin Sure. The relevant reported issue was is here: https://github.com/codeclimate/codeclimate-rubocop/issues/39#issuecomment-184412524.

I reproduced locally in `engines-test` with the following config:

#### .rubocop.yml
```yaml
inherit_from: .rubocop_todo.yml

AllCops:
  DisabledByDefault: true

Lint/UselessAssignment:
  Description: 'Checks for useless assignment to a local variable.'
  StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#underscore-unused-vars'
  Enabled: true
```

#### .rubocop_todo.yml
```yaml
Lint/UselessAssignment:
  Exclude:
    - baby.rb
```

Running `codeclimate analyze -e rubocop` was reporting several useless assignment violations in `baby.rb` (wrong). Running `rubocop` directly was not (correct). I tracked it down to the fact that when the paths were emitted from `FileListResolver` as absolute paths, it worked (this would be the case, for example, if your `.codeclimate.yml` had no `exclude_paths`), but if the workspace paths from the CLI resulted in relative paths, stuff broke. My statements about Rubocop's internal behavior above come from me diving into that code in an attempt to debug this.

wfleming:
Got a spec working. Had to tweak the engine config for that one since this was partially as a result of how workspace paths were handled.

lib/cc/engine/file_list_resolver.rb
wfleming: @jpignata @GordonDiggs refactored this a bit.

jpignata:
:shipit:

GordonDiggs:
:clap: LGTM

PR #53, opened 2016-02-16
Drop support for "exclude_paths" in engine config

wfleming:
Engine-visible `exclude_paths` have been deprecated for a while, and
we've been removing support from a bunch of engines. Engines should now
rely on `include_paths`, which the CLI will calculate based on the
`exclude_paths` in `.codeclimate.yml`.

@codeclimate/review

### DIFF
--- a/lib/cc/engine/file_list_resolver.rb
+++ b/lib/cc/engine/file_list_resolver.rb
@@ -3,32 +3,11 @@ module CC
     class FileListResolver
       def initialize(root:, engine_config: {}, config_store:)
         @root = root
-        @exclude_paths = engine_config["exclude_paths"] || []
-        @include_paths = engine_config["include_paths"]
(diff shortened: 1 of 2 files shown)


ABaldwinHunter:
LGTM.

Eventually I think we should fix the method that determines whether a file can be analyzed to be more inclusive, (or better leverage rubocop's file walking) but I that's outside the scope here.

(E.g. we check for `.rb` extension, while the [original tool](https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/target_finder.rb#L79) also checks if it's a ruby executable.

ABaldwinHunter:
hm, or maybe that's what we're checking already [here](https://github.com/codeclimate/codeclimate-rubocop/blob/will/drop-exclude-paths/lib/cc/engine/file_list_resolver.rb#L31)

NVM. _bushes_

PR #54, opened 2016-02-16
rake docs: pull currently used rubocop version

wfleming:
We were hardcoding this tag, which of course led to it being forgotten
for an update when we updated rubocop.

Rubocop tags their repository reliably, so I think it's reasonable to
use the loaded rubocop's version here to decide what tag to checkout.

@codeclimate/review

### DIFF
--- a/lib/tasks/docs.rake
+++ b/lib/tasks/docs.rake
@@ -1,7 +1,8 @@
 namespace :docs do
   desc "Scrapes documentation from the rubocop gem"
   task :scrape do
-    TAG = "v0.34.2"
+    require "rubocop"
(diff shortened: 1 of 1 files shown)


jpignata:
LGTM. Thanks for the catch!

wfleming:
@jpignata and good thing we had a test for those doc contents: it caught a problem caused by some changes to Rubocop's sources. Fixed in ea0b161 if you want to take a look. Not exactly the cleanest code: I can refactor tomorrow if you'd like.

jpignata:
Closing since I made a duplicative fix in https://github.com/codeclimate/codeclimate-rubocop/pull/57. Oops.

PR #56, opened 2016-02-17
FileListResolver: handle non-existent files

wfleming:
A couple crashes came up this morning as a result of non-existent files trying to be read by the file list resolver. This bug surfaced because of the changes to trying to make all paths absolute.

Took the opportunity to test this class more broadly, as well.

@codeclimate/review


GordonDiggs:
Nice! LGTM

PR #57, opened 2016-02-20
Commit content from cops

jpignata:
Directly commits content from cop class documentation that at least
provides an example or a description that isn't just a restatement
of the check name or description.

The motivation behind this change is to ensure that the content being
exposed through the Code Climate service is useful and gives the user
more details on how to resolve the issue being reported rather than
restating a portion of the description. By directly committing the
content this allows us to start supplementing the content from other
sources such as the Ruby styleguide and our own original content.

@codeclimate/review


wfleming:
Is this effectively just the results of the scrape job committed? I liked having it be an automatic step, since having this committed makes me worry it's  going to get outdated & not updated. What advantages does this give us?

jpignata:
@wfleming This is the results of a subset of the scrape job committed. The intent of this change is to eliminate content that isn't helpful to the end user. The class comments in cops are often thorough descriptions of the check with examples but sometimes are just summaries of the issue description. Our goal is to reduce the amount of exposed content that doesn't provide much value to the end user. By directly committed these contents, it gives us a base on which we can expand the content to be more valuable by supplementing it with other secondary sources and writing our own original content.

PR #61, opened 2016-02-29
Upgrade parser gem

jpignata:
Parser has a bug with out of range characters that was fixed in
2.3.0.5. This fix sets the version requirement for parser to be
~> 2.3.0.5.

See: https://github.com/bbatsov/rubocop/issues/2833

@codeclimate/review


pbrisbin:
LGTM

PR #64, opened 2016-04-18
Explicitly check RuboCop excludes

jpignata:
RuboCop's ConfigStore will sometimes return conflicting answers for
file_to_exclude and file_to_include so check file_to_exclude explicitly
and skip the file in question if it returns true.

c/ @codeclimate/review

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -18,7 +18,9 @@ Style/PercentLiteralDelimiters:
     '%W': '[]'
 Style/StringLiterals:
   Enabled: false
-Style/TrailingComma:
+Style/TrailingCommaInLiteral:
+  Enabled: false
+Style/TrailingCommaInArguments:
   Enabled: false
 Style/DotPosition:
   EnforcedStyle: 'trailing'
--- a/lib/cc/engine/file_list_resolver.rb
+++ b/lib/cc/engine/file_list_resolver.rb
@@ -19,6 +19,8 @@ module CC
 
       private
 
+      attr_reader :config_store
+
       def absolute_include_paths
         @include_paths.map do |path|
           begin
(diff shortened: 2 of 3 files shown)


lib/cc/engine/file_list_resolver.rb
```diff
@@ -30,12 +32,12 @@ def absolute_include_paths
...
-        if file =~ /\.rb$/
-          true
-        else
-          root, basename = File.split(file)
-          @config_store.for(root).file_to_include?(basename)
-        end
+        root, basename = File.split(file)
+        store = config_store.for(root)
+
+        return false if store.file_to_exclude?(basename)
```
GordonDiggs: WDYT of formatting these lines

```ruby
!store.file_to_exclude?(basename) &&
  (file =~ /\.rb$/ || store.file_to_include?(basename))
```

jpignata: I don't love it. With a negation and both `&&` and `||` involved in the expression, I prefer to break it apart.

GordonDiggs:
LGTM

.rubocop.yml
```diff
@@ -18,7 +18,9 @@ Style/PercentLiteralDelimiters:
     '%W': '[]'
 Style/StringLiterals:
   Enabled: false
-Style/TrailingComma:
+Style/TrailingCommaInLiteral:
+  Enabled: false
+Style/TrailingCommaInArguments:
```
ABaldwinHunter: 💄 what do you think about alphabetizing these?

jpignata: Ah, sorry, missed this before merging. The file isn't alphabetized
currently – I'll leave that for the next changeset.

ABaldwinHunter: sounds good

PR #65, opened 2016-05-03
Add path-based fingerprint override for method/class offenses

dblandin:
This PR implements override fingerprints for method and class-based offenses, incorporating the path, check name, and offense message.

@codeclimate/review :mag_right:

### DIFF
--- a/bin/codeclimate-rubocop
+++ b/bin/codeclimate-rubocop
@@ -1,5 +1,5 @@
 #!/usr/bin/env ruby
-$LOAD_PATH.unshift(
+$:.unshift(
   File.expand_path(File.join(File.dirname(__FILE__), "../lib"))
 )
 
--- a/lib/cc/engine/fingerprint.rb
+++ b/lib/cc/engine/fingerprint.rb
@@ -1,0 +1,40 @@
+require "digest/md5"
+
+module CC
+  module Engine
+    class Fingerprint
+      OVERRIDE_COP_NAMES = %w[
+        Metrics/AbcSize
+        Metrics/ClassLength
+        Metrics/CyclomaticComplexity
+        Metrics/MethodLength
+        Metrics/ModuleLength
+        Metrics/PerceivedComplexity
+      ].freeze
+
+      def initialize(path, cop_name, message)
+        @path = path
+        @cop_name = cop_name
+        @message = message
+      end
+
+      def compute
+        if OVERRIDE_COP_NAMES.include?(cop_name)
+          md5 = Digest::MD5.new
+          md5 << path
+          md5 << cop_name
+          md5 << stripped_message
+          md5.hexdigest
+        end
+      end
+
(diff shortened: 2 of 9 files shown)


wfleming:
Do I understand correctly that these fingerprints would be identical to the current default?

I'd like to dig a bit more into the monkey patching before we bail entirely on that. Maybe a good thing to pair on.

dblandin:
> Do I understand correctly that these fingerprints would be identical to the current default?

That's correct.

> I'd like to dig a bit more into the monkey patching before we bail entirely on that. Maybe a good thing to pair on.

Sounds good to me!

dblandin:
@wfleming Ready for another :mag_right:

bin/codeclimate-rubocop
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env ruby
-$LOAD_PATH.unshift(
+$:.unshift(
```
wfleming: Unrelated change? Doid rubocop start complaining about this?

dblandin: Yep, this was recommended by RuboCop in this [rule](http://www.rubydoc.info/github/bbatsov/rubocop/Rubocop/Cop/Style/SpecialGlobalVars).

spec/cc/engine/issue_spec.rb
```diff
@@ -36,6 +36,14 @@ module CC::Engine
           "This cop checks that the cyclomatic complexity of methods is not higher than the configured maximum."
         )
       end
+
+      it "sets a fingerprint for method/class offenses" do
```
wfleming: Worth asserting that it does *not* set a fingerprint for some other cop it shouldn't?

bin/codeclimate-rubocop
wfleming: Oh, it's the new base rubocop. nevermind.

lib/cc/engine/fingerprint.rb
```diff
@@ -0,0 +1,44 @@
...
+          md5 << stripped_message
+          md5.hexdigest
+        end
+      end
+
+      private
+
+      attr_reader :path, :cop_name, :message
+
+      def check_name
```
wfleming: nit, but there's small duplication between this and `Issue#check_name`. And this one doesn't really need the `Rubocop/` bit, since we're just hashing it. Thoughts on using `cop_name` directly in  the hash instead here?

spec/cc/engine/issue_spec.rb
dblandin: Sounds good. Added 👍

lib/cc/engine/fingerprint.rb
dblandin: Sounds good to me. Updated 👍

wfleming:
LGTM

PR #68, opened 2016-07-11
Use Config#target_ruby_version

pbrisbin:
This is the logic that we're replacing, plus smart reading of .ruby-version if
present and not overridden by the AllCops rule.

Required updating rubocop to [0.41.2](https://github.com/bbatsov/rubocop/blob/master/CHANGELOG.md#0412-2016-07-07)

/cc @codeclimate/review can I get a second set of eyes down the rubocop
CHANGELOG to see if I missed any compatibility concerns?

TODO: update docs.codeclimate.com

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,11 +1,11 @@
 source 'https://rubygems.org'
 
 gem "activesupport", require: false
-gem "rubocop", "~> 0.37.2", require: false
+gem "rubocop", "~> 0.41.2", require: false
 gem "rubocop-rspec", require: false
 gem "safe_yaml"
 gem "pry", require: false
-gem "parser", "~> 2.3.0.5"
+gem "parser", "~> 2.3.1.1"
 
 group :test do
   gem "rake"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -7,14 +7,14 @@ GEM
       minitest (~> 5.1)
       thread_safe (~> 0.3, >= 0.3.4)
       tzinfo (~> 1.1)
-    ast (2.2.0)
+    ast (2.3.0)
     coderay (1.1.0)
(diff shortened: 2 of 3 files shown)


wfleming:
LGTM. Nothing jumped out at me in RuboCop's changelog, but did we run this new version against a bunch of repos yet to confirm no breaking config or anything?

pbrisbin:
@wfleming i only ran it on app as a sanity check. do you have a script and/or list of oss repos i can run it on? Do you go so far as to compare output to the old version, or just confirm it doesn't blow up?

wfleming:
> Do you have a script and/or list of oss repos I can run it on?

I wish I did. That might be good to compile for the future. Off the top of my head:

1. app
2. this engine
3. https://github.com/rails/rails
4. https://github.com/jekyll/jekyll

Those seem like a decent sample set?

> Do you go so far as to compare output to the old version, or just confirm it doesn't blow up?

I would just want to confirm it doesn't blow up.

pbrisbin:
Wrote up a script to run over those projects, I think it's a good list. Since I was scripting anyway I added running the old engine and diffing the JSON outputs anyway.

pbrisbin:
Results:

- app: many new cops triggering, some existing issues changing because of smarter source locations
- rubocop engine: it fixed not knowing the target ruby version (because it now reads .ruby-version)
- rails: it started emitting "use new HashSyntax" issue. configured TargetRubyVersion is 2.3 so it should've been emitting these all along IMO
- jekyll: old engine emitted 1 conditional-assignment issue, new engine emits nothing. behavior around that cop has changed to be smarter so this is likely a Good Thing

All in all, I think it's a win and I should move forward. WDYT @wfleming ?

wfleming:
I agree.

PR #73, opened 2016-08-23
Correctly strip detailed message in fingerprint

ABaldwinHunter:
A user may elect more detailed messaging in RuboCop output by configuring:

    AllCops:
      DisplayCopNames: true
      DisplayStyleGuide: true
      ExtraDetails: true

which adds a url to issue messages, part of the fingerprint compute.

To correctly compute fingerprints in this case, we need an extra strip step.

@codeclimate/review 

cc @almostwhitehat with a thanks for finding this bug.

### DIFF
--- a/lib/cc/engine/fingerprint.rb
+++ b/lib/cc/engine/fingerprint.rb
@@ -12,6 +12,9 @@ module CC
         Metrics/PerceivedComplexity
       ].freeze
 
+      URL_REGEX = / \(https?\:.+\)/
+      LINES_REGEX = / \[.+\]$/
+
       def initialize(path, cop_name, message)
         @path = path
         @cop_name = cop_name
@@ -33,7 +36,7 @@ module CC
       attr_reader :path, :cop_name, :message
 
       def stripped_message
-        message.gsub(/ \[.+\]$/, "")
+        message.gsub(URL_REGEX, "").strip.gsub(LINES_REGEX, "")
       end
     end
   end
--- a/spec/cc/engine/fingerprint_spec.rb
+++ b/spec/cc/engine/fingerprint_spec.rb
@@ -1,0 +1,54 @@
+require "spec_helper"
+require "cc/engine/fingerprint"
+
+module CC::Engine
+  describe Fingerprint do
+    describe "#compute" do
+      it "returns a fingerprint for method/class offenses" do
+        cop_name = "Metrics/ModuleLength"
+        path = "foo/bar.rb"
+        message = "foo"
+
+        computed = Fingerprint.new(path, cop_name, message).compute
+
+        expect(computed).to eq "53a5f182884e58dcb16d24fca37e10bc"
+      end
+
+      it "returns nil for non method/class offenses" do
+        cop_name = "Style/Foo"
+        path = "foo/bar.rb"
+        message = "foo"
+
+        computed = Fingerprint.new(path, cop_name, message).compute
+
+        expect(computed).not_to be
+      end
+
+      it "computes same fingerprint regardles of message url detail" do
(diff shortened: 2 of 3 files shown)


lib/cc/engine/fingerprint.rb
```diff
@@ -33,7 +33,13 @@ def compute
       attr_reader :path, :cop_name, :message
 
       def stripped_message
-        message.gsub(/ \[.+\]$/, "")
+        # example full message with details:
+        # Metrics/MethodLength: Method has too many lines. [18/10] (https://github.com/bbatsov/ruby-style-guide#short-methods)
+        url_regex = %r{ \(https\://github\.com.+\)}
```
GordonDiggs: Is it always definitely a github url?

ABaldwinHunter: In all of the examples I saw it points to 
https://github.com/bbatsov/ruby-style-guide

Double checking the RuboCop source code.

almostwhitehat: Not always - the url can be overridden with the StyleGuide and Reference keys: https://github.com/bbatsov/rubocop/blob/da01214a7de225dd62ebe2df9ae54d56d7a43e54/lib/rubocop/cop/cop.rb#L226:L234

ABaldwinHunter: Looks like it's an issue attribute that comes from
https://github.com/bbatsov/rubocop/blob/master/config/enabled.yml#L5

I don't think it would hurt to make the regex more aggressive though. I can't think of any false positives involving urls in description.

ABaldwinHunter: @almostwhitehat Ah good call, thanks! missed your message before. Updating

ABaldwinHunter:
@GordonDiggs Ready for another look!

lib/cc/engine/fingerprint.rb
```diff
@@ -33,7 +33,13 @@ def compute
       attr_reader :path, :cop_name, :message
 
       def stripped_message
-        message.gsub(/ \[.+\]$/, "")
+        # example full message with details:
+        # Metrics/MethodLength: Method has too many lines. [18/10]  (https://github.com/bbatsov/ruby-style-guide#short-methods)
+        url_regex = / \(http[s]?:.+\)/
+        lines_regex = / \[.+\]$/
```
pbrisbin: Should these be constants?

lib/cc/engine/fingerprint.rb
```diff
@@ -33,7 +33,12 @@ def compute
       attr_reader :path, :cop_name, :message
 
       def stripped_message
-        message.gsub(/ \[.+\]$/, "")
+        # example full message with details:
+        # Metrics/MethodLength: Method has too many lines. [18/10]  (https://github.com/bbatsov/ruby-style-guide#short-methods)
```
pbrisbin: I think the spec is enough and don't think this comment is hugely valuable. WDYT?

spec/cc/engine/fingerprint_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  describe Fingerprint do
+    describe "#compute" do
+      it "returns a fingerprint for method/class offenses" do
+        cop_name = "Metrics/ModuleLength"
+        path = "foo/bar.rb"
+        message = "foo"
+
+        computed = Fingerprint.new(path, cop_name, message).compute
+
+        expect(computed).to eq("53a5f182884e58dcb16d24fca37e10bc")
```
pbrisbin: 💄 we typically don't use parens with `eq`: `expect(x).to eq y`

GordonDiggs:
LGTM

PR #74, opened 2016-09-26
Add RuboCop Migrations

magec:
Hi!

We want to start using a new rubocop extension we developed (https://rubygems.org/gems/rubocop-migrations). It is a simple gem that disallows index removal from rails migrations (so you don't accidentally remove an index that is being used).

Given that we use codeclimate, it would be nice to be able to use that there.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@ source 'https://rubygems.org'
 gem "activesupport", require: false
 gem "rubocop", "~> 0.41.2", require: false
 gem "rubocop-rspec", require: false
+gem "rubocop-migrations", require: false
 gem "safe_yaml"
 gem "pry", require: false
 gem "parser", "~> 2.3.1.1"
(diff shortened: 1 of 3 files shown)


ivanilves:
:+1:

enricostano:
:+1:

dblandin:
Hey @magec,

This looks good to me. You might want to also adjust the [category mapping](https://github.com/codeclimate/codeclimate-rubocop/blob/master/lib/cc/engine/category_parser.rb), otherwise this cop will emit `Style` issues which doesn't seem appropriate. You can see all supported categories [here](https://github.com/codeclimate/spec/blob/master/SPEC.md#categories). Maybe `Performance` would make sense?

magec:
Thanks for reviewing @dblandin !

Added the category.

dblandin:
Hey @magec,

Just wanted to follow-up and let you know that I've released an engine update the includes the new rubocop plugin.

Let me know if you run into any issues!

Thanks again!

PR #75, opened 2016-10-12
Support automatic channel building for rubocop

wfleming:
This moves Rubocop to use Patrick & support automatic channel deployments based on branch name


wfleming:
cc @codeclimate/review

PR #78, opened 2016-10-13
Add GEM_PATH env var to Dockerfile

nporteschaikin:
...to use code's bundle, which may include rubocop extensions.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,5 +16,6 @@ USER app
 
 VOLUME /code
 WORKDIR /code
+ENV GEM_PATH /code/vendor/bundle/ruby/2.2.0
 
 CMD ["/usr/src/app/bin/codeclimate-rubocop"]


wfleming:
Doesn't setting this `ENV` var stop loading gems from the standard location?

dblandin:
> Doesn't setting this ENV var stop loading gems from the standard location?

Nope. `GEM_PATH` adds entries in addition to the standard system location.

```
$ gem environment
  ...
  - GEM PATHS:
     - /usr/lib/ruby/gems/2.2.0
     - /home/app/.gem/ruby/2.2.0
  ...
$ export GEM_PATH=/code/vendor/bundle/ruby/2.2.0
$ gem environment
  ...
  - GEM PATHS:
     - /usr/lib/ruby/gems/2.2.0
     - /code/vendor/bundle/ruby/2.2.0
  ...
```

wfleming:
Cool, my bad. In that case LGTM, I think.

PR #80, opened 2016-10-14
Override default fingerprinting for Metrics/BlockLength

maxjacobson:
Noticed in a recent build that this cop is showing up as fixed/new
rather than changed. It was added in RuboCop v0.44, which we just
upgraded to.

@codeclimate/review

### DIFF
--- a/lib/cc/engine/fingerprint.rb
+++ b/lib/cc/engine/fingerprint.rb
@@ -5,6 +5,7 @@ module CC
     class Fingerprint
       OVERRIDE_COP_NAMES = %w[
         Metrics/AbcSize
+        Metrics/BlockLength
         Metrics/ClassLength
         Metrics/CyclomaticComplexity
         Metrics/MethodLength


wfleming:
Code seems fine, but the PR description seems like the opposite of what this is doing? This opts that cop *out* of default fingerprinting.

maxjacobson:
@wfleming yea, good call. I think what I meant was that, with this change, we would neglect to provide a fingerprint, which makes the system choose a fingerprint using the default fingerprinting algorithm, which overrides this particular engine's behavior.. so I think 2 ways of saying the same thing? 😅

dblandin:
> I think what I meant was that, with this change, we would neglect to provide a fingerprint

@maxjacobson I think that's also the opposite of what this PR accomplishes. By including `Metrics/BlockLength` within `OVERRIDE_COP_NAMES`, the engine computes and emits an override fingerprint consisting of file path, cop name, and issue message.

If `Metrics/BlockLength` were not included within `OVERRIDE_COP_NAMES`, the issue would adopt the default fingerprint (computed [here](https://github.com/codeclimate/codeclimate/blob/master/lib/cc/analyzer/source_fingerprint.rb)) consisting of file path, check name, and source content. The default fingerprinting approach doesn't work well for issues like complexity, method length, or, in this case, block length since the source content can easily change without eliminating the issue.

dblandin:
Anyway, I think the disagreement is in semantics. I think we're practically on the same page. :smile:

PR #81, opened 2016-10-17
Clean up Makefile, release to docker hub

pbrisbin:
- Remove IMAGE_NAME

  This is an implementation detail of the Makefile. With the new release
  task, the registry and tag can be injected, but the engine name is
  fixed (to prevent errors) and the image name as built (and used for
  tests) is private (to avoid confusion).

- Add a test target

  This is tailored for CI usage more than local, since that's by far the
  most frequent use-case.

- Add a release target

  By injecting the registry and tag this is re-usable for GCR. Even
  though that's temporary, the flexibility is still valuable going
  forward.

  The caller is expected to have already logged in at the location
  they're releasing two. This helps keep the Makefile interface simple.

- Release to DockerHub

  This is where engine images will reside long-term. We continue to
  release to GCR additionally to not break production during the
  transition.

/cc @codeclimate/review

Now's the time to bikeshed! Once everyone's good on the general pattern of
Makefile+circle.yml here, I'm going to implement it across all the engines.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,17 @@
-.PHONY: image test
+.PHONY: image test release
 
-IMAGE_NAME ?= codeclimate/codeclimate-rubocop
+RELEASE_REGISTRY ?= codeclimate
+RELEASE_TAG ?= latest
 
 image:
-	docker build --rm -t $(IMAGE_NAME) .
+	docker build --tag codeclimate/codeclimate-rubocop .
 
-test: image
-	docker run --rm $(IMAGE_NAME) sh -c "cd /usr/src/app && bundle exec rake"
+test:
+	docker run --workdir /usr/src/app \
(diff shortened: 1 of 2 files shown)


wfleming:
Typo in commit message under "Add a release target": you have "two", and mean "to".

I'm cool with the general pattern. I think you could extract some repetition like `IMAGE_NAME` to `IMAGE_NAME=rubocop` (without the `?` so it can't be overridden) to reduce a bit of duplication/typo risk, but maybe that's not worth it/won't improve readability.

pbrisbin:
Thanks @wfleming I'm actually going to close this for now since we're going to just handle it in Shipbot out of hackweek-expediency. I'll come back when we action actually cleaning up individual engines' CI scripts.

PR #83, opened 2016-11-01
upgrade to RuboCop 0.45.0

wfleming:
This was released yesterday.

Compare: https://github.com/bbatsov/rubocop/compare/v0.44.1...v0.45.0

cc @codeclimate/review


maxjacobson:
1. the spec I merged will prob require you to update the whitelist or provide readups for new cops
2. my earlier merge failed to deploy and I haven't fixed it yet bc I've gotten distracted looking into a new escalation

If you're planning to add readups for any new cops, I'll wait to look at those before 👍-ing

maxjacobson:
Wondering if the security cop is not scraper-compatible because it's not documented, or because it's a net cop category so the folder doesn't exist or something?

wfleming:
I've re-scraped docs and checked them in: this moves away from what we did back in February: https://github.com/codeclimate/codeclimate-rubocop/commit/04c8e566731c521f8b31385e9f5cfd42e5ebbd46.

As time has gone on, we've basically done a poor job of writing new docs, while the RuboCop team has done a very good job: writing docs for new cops, as well as improving/updating docs for existing ones.

I think we should include this content again. If we still see cops that we don't want to include docs from for whatever reason, I think we should specifically exclude them from future updates. @codeclimate/review

GordonDiggs:
@wfleming should we start scraping as part of the build again?

GordonDiggs:
LGTM whatever you want to do re: scraping

wfleming:
@GordonDiggs I think I'm going to defer re-adding that to the build. There still might be some value in having the opportunity to, for example, edit docs ourselves & commit them. Since we have the spec for checking for missing docs now, I think this re-add is a reasonable first step.

@maxjacobson good call about security missing: I've added that.

Both of you: I took a quick stab at excluding "bad" docs with a simple heuristic in the rake task. More review, please!

GordonDiggs:
I like it! LGTM

maxjacobson:
WDYT about adding a section to the README about the process for updating rubocop, including re-running this rake task? The spec will catch some things, but it won't include, e.g., typo fixes to docs

wfleming:
@maxjacobson Added a quick `DEVELOPERS.md`

maxjacobson:
Sweet, this looks great to me

PR #87, opened 2016-12-21
Bump dependencies

pointlessone:
* Switch to official docker Ruby image
* Bump Ruby to 2.3.3
* Bump Rubocop to 0.46.0
* Bump other gems as well


spec/support/currently_undocumented_cops.txt
```diff
@@ -1,3 +1,5 @@
+RuboCop::Cop::Bundler::DuplicatedGem
```
maxjacobson: It looks like these new cops are documented in the rubocop codebase -- any idea why the scraper wasn't able to pull them in?

spec/support/currently_undocumented_cops.txt
```diff
@@ -1,3 +1,5 @@
...
+RuboCop::Cop::Bundler::OrderedGems
```
maxjacobson: the introduction of a new namespace should be fine - [by default these will be categorized as "style".](https://github.com/codeclimate/codeclimate-rubocop/blob/master/lib/cc/engine/category_parser.rb) I think that probably sounds right... WDYT?

maxjacobson:
Looking into why the tests didn't run for this PR

maxjacobson:
`Build forked pull requests` was set to off -- I believe it's because this repo was private and was later open source. It's fine to set that to on. Will you push again to trigger a build?

spec/support/currently_undocumented_cops.txt
pointlessone: TBH, I just forgot to run the scrapper task. Fixed now.

pointlessone:
Yes, the build failed because of base image change. Weird enough, the engine runs just fine it's just the way specs are invoked revealed the issue.

Updated the `Docekrfile`.

maxjacobson:
Cool, fix looks good.

The one remaining thought I have is that, while we're updating this engine, I think it might be good to check our default `.rubocop.yml` configuration. We disable all cops, and then enable _specific_ cops that we think are going to be useful to the most number of people. Do you think any of these new cops fit that criteria?

Dockerfile
```diff
@@ -1,12 +1,12 @@
-FROM codeclimate/alpine-ruby:b38
+FROM ruby:2.3-alpine
```
GordonDiggs: One thing we're conscious of is the size of the images. What does this do to the size of this engine's Docker image?

pointlessone: Current is 69.7 MiB. The new one is 165.9 MiB. So yes, increase of almost 100 MiB.
What is the threshold?

pointlessone:
I personally try keeping my Rubocop config very slim and use community-driven default style.

But If I had to chose I'd probably enabled `Bundler/DuplicatedGem`, `Performance/CompareWithBlock`, `Style/EmptyMethod`, and `Style/TernaryParentheses`.

Dockerfile
GordonDiggs: Our spec says [512 MB](https://github.com/codeclimate/spec/blob/master/SPEC.md#resource-restrictions) so this should be ok!

maxjacobson:
SGTM. That default config lives [here](https://github.com/codeclimate/codeclimate/blob/master/config/rubocop/.rubocop.yml). We can make that change separately.

Gemfile.lock
```diff
@@ -1,72 +1,71 @@
...
       parser (>= 2.3.1.1, < 3.0)
       powerpack (~> 0.1)
       rainbow (>= 1.99.1, < 3.0)
       ruby-progressbar (~> 1.7)
       unicode-display_width (~> 1.0, >= 1.0.1)
     rubocop-migrations (0.1.2)
       rubocop (~> 0.41)
-    rubocop-rspec (1.5.0)
-      rubocop (>= 0.40.0)
+    rubocop-rspec (1.8.0)
```
andyw8: Thanks, I was just about to open a PR for this! (I'm one of the rubocop-rspec maintainers).

PR #91, opened 2017-02-14
Upgrade to RuboCop 0.47.1 (again)

maxjacobson:
This needs to be shipped in tandem with https://github.com/codeclimate/codeclimate/pull/563

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,8 @@ WORKDIR /usr/src/app
 COPY Gemfile /usr/src/app/
 COPY Gemfile.lock /usr/src/app/
 
-RUN gem install bundler && \
+RUN gem update --system && \
+    gem install bundler && \
     bundle install -j 4 && \
     rm -fr /usr/share/ri
 
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@ source 'https://rubygems.org'
 gem "activesupport", require: false
 gem "parser", "~> 2.3.3.1"
 gem "pry", require: false
-gem "rubocop", "~> 0.45", require: false
+gem "rubocop", "~> 0.47.1", require: false
 gem "rubocop-migrations", require: false
 gem "rubocop-rspec", require: false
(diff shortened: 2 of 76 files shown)


raulmarcosl:
Please, be careful with this upgrades. Yesterday our builds were failing due to cop naming problems, so we upgrade our rubocop config to `0.47.1`, but then you reverted the changes, and our new config is failing, so we had to revert it.

IMO, it would be great to be able to force a rubocop version to avoid this problems.

tarebyte:
Friendly bump, would be lovely to have this merged in 😄

colszowka:
Make that 0.48.1 please :)

In that version Rubocop changed `Rubocop/Style/PercentLiteralDelimiters` to be `[]` by default, which broke our code climate reporting (on the bundled 0.46) while locally everything's green on the fresh version. Additionally, the bundled 0.46 version does not seem to acknowledge the `PreferredDelimiters: / default: "[]"` setting, so we had to disable the cop in code climate for the time being.

trautwein:
Another friendly bump 😄  Any update on this?

kjg:
With the vast number of differences between rubocop versions, it would be awesome if we could somehow specific which version we'd like codeclimate to run (to match our local gem version)

deivid-rodriguez:
Hei! Not sure what's blocking this, but there's this [tool](https://github.com/pocke/mry) to migrate rubocop's config between versions. Just mentioning in case it could help.

maxjacobson:
Just an update:

Thanks everyone for your patience on this.

If you're interested in using RuboCop 0.48.1 on Code Climate, you can now. We've made it available as a channel. You can switch to it by adding the following to your `.codeclimate.yml`:

```yaml
engines:
  rubocop:
    enabled: true
    channel: rubocop-0-48
```

The default channel is still running RuboCop 0.46.0 as of today.

I'm going to close this PR for now.

(btw, thanks for sharing mry, @deivid-rodriguez, that's pretty neat and may help us out in the future)

PR #95, opened 2017-05-12
Expose docs generation through make

filipesperandio:


Makefile
```diff
@@ -2,6 +2,12 @@
 
 IMAGE_NAME ?= codeclimate/codeclimate-rubocop
 
+docs:
```
maxjacobson: I think...

1. this should depend on image (`docs: image`)
1. this should be added to the `PHONY` list at the top of the file
1. this shouldn't be the first task -- by convention, we always have `image` be the first task, so `make` will default to `make image`.

filipesperandio:
The error is weird, went smooth for me.

filipesperandio:
@maxjacobson Could it be some local permission not allowing the user within docker to write to the directory?

maxjacobson:
@filipesperandio yeah I think so. Possibly a linux/mac difference.

@wfleming perhaps you know the docker magic invocation to make this work?

wfleming:
It's running as user id `9000` which doesn't exist on the host machine & doesn't have write permissions. My guess is it works for Filipe because of some weird Docker for Mac host-fs-virtualization magic to handle permissions between the host/the VM that actually hosts docker. I think the easy fix is to pass `--user root` to the `docker run`. (which might lead to people needing to `sudo chown` locally for some things, but we do that in other cases and it's fine.)

maxjacobson:
Cool, I can confirm that adding `--user root` helps for me on linux. Do things still work for you with that change @filipesperandio?

maxjacobson:
(Thanks @wfleming!)

filipesperandio:
@maxjacobson @wfleming That works on mac too! 👍

PR #101, opened 2017-08-23
RuboCop 0.49.1

### LINKED ISSUE #99: bump default version to 0.49.1 (or greater)
There is an open CVE for rubocop versions less than 0.49.1.

see https://github.com/bbatsov/rubocop/issues/4336

pointlessone:
Notable changes:

* RuboCop 0.49.1
* Ruby 2.4.1
* Updated cop docs
* Smaller image (170MB → 77.2MB)

Resolves #99


maxjacobson:
I think we need to be careful about replacing the default channel with `0.49.1`, because that will upgrade all users who are currently using `0.46.0`, which is likely to cause errored builds due to incompatible configuration.

I'd like to propose:

1. Let's introduce a `channel/rubocop-0-46` branch which is based on whatever is currently on master (and make this channel available on prod and CLI)
2. Make a branch `channel/rubocop-0-49` which is whatever is currently on master
3. Update this PR to target that branch
4. Get this PR reviewed and merged into `channel/rubocop-0-49`, release `0.49.1` to prod and CLI
5. Offer to people who would like to use `0.49.1` that they can now switch their channel
6. (At some point, not 100% sure when) merge `channel/rubocop-0-49` into `master` and update [the default rubocop config](https://github.com/codeclimate/codeclimate/blob/master/config/rubocop/.rubocop.yml) to be `0.49.1`-compatible and when people have issues due to checked-in configuration, be prepared to explain to them that they can lock to `0.46` which will continue to work, or be prepared to help them fix their configuration

pointlessone:
@maxjacobson 1—3 are done.

Dockerfile
```diff
@@ -1,17 +1,15 @@
...
-COPY Gemfile /usr/src/app/
-COPY Gemfile.lock /usr/src/app/
 
-RUN gem install bundler && \
-    bundle install -j 4 && \
-    rm -fr /usr/share/ri
+COPY Gemfile Gemfile.lock ./
 
-RUN apk --update add git
+RUN gem install --no-document -g
```
dblandin: Does this replace the `bundle install` command?

dblandin:
> Smaller image (170MB → 77.2MB)

Awesome :+1:

Dockerfile
pointlessone: It totally does. Specifically, the `-g` option.

dblandin: ~~Nice. Could we expand that option to its full form? `--file`. Weird that `-g` is the shortform. I'd assume it would be short for `--global`.~~

Looks like Gemfile support within the `gem` command has some issues and is not  yet "production-ready": https://github.com/rubygems/rubygems/issues/1207#issuecomment-86158101

> It is possible to use the Gemfile support. It is recommended and currently supported to use Bundler instead.

In particular, it looks like there are some known security vulnerabilities, like this CVE: https://github.com/rubygems/rubygems/issues/1211

Sounds like we should stick with `bundle install` for now.

pointlessone: The comment if over 2 years old. RubyGems has come a long way since then. There's a bit of feature parity issue (missing features). But it's pretty robust when it comes to installing gems from a lockfile.

The [CVE](https://access.redhat.com/security/cve/CVE-2013-0334) originally affects Bundler. By the nature of the issue (source order affecting gem lookup) it affects RubyGems in the same way. The issue is not resolved (resolvable?) both in RubyGems and Bundler.

dblandin: > The comment if over 2 years old. RubyGems has come a long way since then. There's a bit of feature parity issue (missing features). But it's pretty robust when it comes to installing gems from a lockfile.

Two years isn't so long in the open source world :)

> The CVE originally affects Bundler. By the nature of the issue (source order affecting gem lookup) it affects RubyGems in the same way. The issue is not resolved (resolvable?) both in RubyGems and Bundler.

Based on the [Bundler blog post](http://bundler.io/blog/2014/08/14/bundler-may-install-gems-from-a-different-source-than-expected-cve-2013-0334.html), didn't Bundler 1.7.0 address that specific CVE? I don't see a version of `gem` that claims a similar fix. If you can refer me to some documentation that recommends using gem `--file/-g` vs bundle install, I'm happy to give it a try, but all I've read so far seems to discourage its use and instead rely on `bundle install`, so I'm leaning towards sticking with that.

pointlessone: Reverted to Bundler since it's easier than finding any formal `gem i -g` endorsements.

However:

* The CVE is not relevant to this particular case. We have only one source.
* The CVE was partially addressed in Rubygems. The proposed bundler fix was implemented except for emitted warning.
* Anecdotal "for": apart from many of my own projects, we have successfully used `gem i -g` in the Grep engine.

pointlessone: Had to add extra clean up but the image is even smaller now: 74.6MB.

PR #112, opened 2017-12-13
Rubocop 0.52.0

pointlessone:
Updated docs are included.


pointlessone:
@chrishulton I also intend to promote it to `stable`. This is a rather safe upgrade from 0.51. 0.51 was tested for a while now.

Would you like to reevaluate your decision? 😄

chrishulton:
@pointlessone Ah, ok.  I think that's reasonable.  I'd suggest trying a few runs of the channel before promoting just as a sanity check, you should be able to do that via CLI.  If that looks good, LGTM.

PR #113, opened 2017-12-15
Switch stable to Rubocop 0.52

pointlessone:
This PR contains no changes from `channel/rubocop-0.52`. It only bumps the `stable` to Rubocop 0.52.


Dockerfile
```diff
@@ -1,18 +1,18 @@
...
-RUN gem install bundler && \
-    bundle install -j 4 && \
-    rm -fr /usr/share/ri
+RUN adduser -u 9000 -D app
 
-RUN apk --update add git
+COPY Gemfile Gemfile.lock /usr/src/app/
+
+RUN gem install bundler && \
+    bundle install -j 4 --without=test && \
```
wfleming: I appreciate that you always look for ways to make our docker images smaller, but I think this is going further than necessary: running `bundle install` every time you run tests locally makes the local feedback cycle much too slow. The test gems really aren't that big, we install them into almost all of our production images, so I'd prefer we keep these.

pointlessone: OK. I'll take a look a bit later at this.

PR #129, opened 2018-07-26
Add `bin/newver` script for creating new Rubocop channels.

wilson:


bin/newver
```diff
@@ -0,0 +1,38 @@
...
+  exit 1
+fi
+NEWVER="$1"
+
+git checkout Gemfile*
+git checkout -b "newver/${NEWVER}"
+
+ed Gemfile <<-EDITS
+/gem "parser",/c
+gem "parser"
```
maxjacobson: is this replacing the line with the same line?

bin/newver
```diff
@@ -0,0 +1,38 @@
...
+.
+/gem "rubocop",/c
+gem "rubocop", "~> ${NEWVER}", require: false
+.
+wq
+EDITS
```
maxjacobson: TIL some `ed` basics, that's neat :+1:

bin/newver
```diff
@@ -0,0 +1,38 @@
...
+
+bundle update
```
maxjacobson: should we do something more scoped like, `bundle update rubocop`?

maxjacobson: wdyt about doing this in docker? In general we try to avoid needing to have non-docker things installed on your host. We could do something like:

```shell
docker run --user root --rm --volume "$(pwd)":/usr/src/app --workdir /usr/src/app codeclimate/codeclimate-rubocop bundle update rubocop
```

In a lot of our code bases, we've added such a task to the Makefile as `make bundle`. If we did that, we might invoke it here as `make bundle -e BUNDLE_ARGS='update rubocop'`

bin/newver
```diff
@@ -0,0 +1,38 @@
...
+make image
+make docs
+sudo chown -R $(id -u):$(id -g) config
+
+set +e
+make test
+retval=$?
+if [ $retval -ne 0 ]; then
+  >&2 echo "If tests fail, you may need to modify spec/support/currently_undocumented_cops.txt"
+  exit $retval
```
maxjacobson: this is a nice CLI UX :clap:

bin/newver
maxjacobson: Note: in order for this to work, it would be necessary to add a `make image` before the `ed` line

bin/newver
wilson: It's replacing the version-specific line on `master` with one that doesn't specify a version. Master is pinned to an oddly-old parser, it seemed to me.

bin/newver
wilson: That seems like a nice improvement; I will change it to invoke docker like your example.

maxjacobson: Whoops, this comment was meant to go on the thread about using docker, not the thread about scoping the bundle update :sweat_smile:

ale7714:
@maxjacobson I updated the script. I think it would be nice to merge this because it helps with releasing new versions. Can i get another review?

PR #155, opened 2019-02-08
Update redundant_self.md

davehenton:
fix typo

### DIFF
--- a/config/contents/style/redundant_self.md
+++ b/config/contents/style/redundant_self.md
@@ -6,7 +6,7 @@ The usage of `self` is only needed when:
     presence of a method name clash with an argument or a local
     variable.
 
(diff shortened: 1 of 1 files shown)


maxjacobson:
:wave: @davehenton. FYI, these files are auto-generated based on [the docs upstream in RuboCop](https://github.com/rubocop-hq/rubocop/blob/7e9bc4f2762f03ba9bb397d9a5f3f0b8cd9f9988/lib/rubocop/cop/style/redundant_self.rb#L14). That means that we _can_ make this change here, but it'll get overwritten next time we update RuboCop to a newer version. What would you think about submitting that fix upstream?

wfleming:
Closing this. As max, said, if we fix it here it'll just get overwritten next time we pull docs from rubocop.

PR #161, opened 2019-03-21
Migrate to circle2

ale7714:
Old branches will need to cherry pick this commit


.circleci/config.yml
```diff
@@ -0,0 +1,41 @@
...
+
+jobs:
+  test:
+    machine:
+      docker_layer_caching: true
+    working_directory: ~/codeclimate/codeclimate-rubocop
+    steps:
+      - checkout
+      - run: make image
+      - run: make citest
```
maxjacobson: per the failing build, it looks like this should be `make test`

ale7714:
Sorry this is WIP, requested review too eagerly. Wrapping up one more thing. And yes to both things

.circleci/config.yml
```diff
@@ -0,0 +1,49 @@
...
+      - run: make test
+
+  release_images:
+    machine:
+      docker_layer_caching: true
+    working_directory: ~/codeclimate/codeclimate-rubocop
+    steps:
+      - checkout
+      - run:
+          name: Validate owner
```
ale7714: circle 1 has this nice feature https://github.com/codeclimate/codeclimate-rubocop/blob/master/circle.yml#L24. 

Unfortunately not available in circle2 so we're the validation within the job itself. 

`circleci step halt` is a nifty trick to exit the job without failing the build. I learn it from this [post](https://discuss.circleci.com/t/useful-tips-and-best-practices-when-migrating-to-circleci-2-0/23075).

maxjacobson: That is neat!

PR #168, opened 2019-05-04
Add Support for RuboCop 0.68

natesholland:
### Summary

This adds support for RuboCop 0.68

One key change I want to call out here is in 32aec57. Previous we were ignoring documentation in files like the following because there was a comment before the main doc comment. This was causing the file to be filtered out because the docs were thought to be too short. I fixed this by adding a look ahead for the class definition.
https://github.com/rubocop-hq/rubocop/blob/2a4f4f0fdac4e1bb25891fc312af2ea60c6aa2f0/lib/rubocop/cop/layout/indent_first_argument.rb

This ended up bringing in docs for a few cops that were previously labeled as undocumented but were actually just formatted differently than expected.

### Logistics

This work will need to be merged into a new branch `channel/rubocop-0-68`, though I do not have the permissions to create this branch.

This is my first time working in this repository so please let me know if there's anything I can change or do to help.


natesholland:
Should I be running `rake docs:scrape`? If I do that the specs fail because `RuboCop::Cop::Layout::IndentFirstArgument` and `RuboCop::Cop::Layout::AlignArguments` lose their documentation. Does this mean I need to add documentation for those two somewhere in rubocop?

natesholland:
I created this PR with the intention of bumping to support 0.68 but ended up creating another change to look ahead in the regex for the class definition. If you would like me to split 32aec57 and 7c05143 apart into their own PR and have two separate PRs I can.

koic:
FYI, please bump RuboCop Performance from 1.1.0 to 1.2.0 to apply `Performance/RegexpMatch` false negative fix when upgrading to RuboCop 0.68.
https://github.com/rubocop-hq/rubocop-performance/pull/47

natesholland:
Thanks for the tip @koic. Done.

I don't know that this needs to block this PR, but should we make a rake task to pull the performance cops from `rubocop-performance` similar to the existing `docs:scrape`?

natesholland:
@wfleming I'd like to bump this since some other efforts are blocked by it.

filipesperandio:
@natesholland Rubocop 0.68 was [added last week](https://github.com/codeclimate/codeclimate-rubocop/pull/173/commits)
I think it would be nice if you can split the part of the docs regex and rebase it off the new branch. I can look at it then.

natesholland:
Thanks @filipesperandio I am closing this in favor of #201 where I have pulled the fix out into its own PR.

PR #180, opened 2019-06-03
Updated channel and documentation to Rubocop v0.71

filipesperandio:


spec/cc/engine/config_upgrader_spec.rb
```diff
@@ -15,6 +15,8 @@ def get_true
       create_source_file(".rubocop.yml", <<~CONFIG)
         Style/AccessorMethodName:
           Enabled: false
+        Style/FrozenStringLiteralComment:
+          Enabled: false
```
filipesperandio: Had to add this for the spec to pass, but not certain if this is the way to go.

PR #190, opened 2019-06-05
Add rubocop-performance back

filipesperandio:


filipesperandio:
Messed this up on rebase and is actually a duplicate from #194

Issue #187, opened 2019-06-04
rubocop-performance missing in 0.68+


granth:
The `rubocop-performance` gem is included in the `rubocop-0-67` channel, but not `rubocop-0-68` and up.

https://github.com/codeclimate/codeclimate-rubocop/blob/c459430a62401a36ed926df7649c4baa983e8a79/Gemfile#L9-L12
https://github.com/codeclimate/codeclimate-rubocop/blob/94f870a1775bb49adfdeebf7f68803cb9ad2433f/Gemfile#L9-L11

filipesperandio:
https://github.com/codeclimate/codeclimate-rubocop/pull/189
https://github.com/codeclimate/codeclimate-rubocop/pull/191
https://github.com/codeclimate/codeclimate-rubocop/pull/192
https://github.com/codeclimate/codeclimate-rubocop/pull/194

PR #191, opened 2019-06-05
Add rubocop-performance back

filipesperandio:


filipesperandio:
RE: #187 and https://github.com/codeclimate-support/escalations/issues/634

PR #193, opened 2019-06-05
Create branch for new channel based on the latest channel instead of master

filipesperandio:


bin/newver
```diff
@@ -7,11 +7,19 @@ if [ $# -eq 0 ]; then
...
+}
+
 git checkout Gemfile*
+git fetch origin
+git checkout $(previous_channel)
 git checkout -b "newver/${NEWVER}"
 
 ed Gemfile <<-EDITS
-/gem "parser",/c
+/gem "parser"/c
```
maxjacobson: can I be honest? I have no idea how to use `ed` or what this change accomplishes. Would you mind providing some context?

bin/newver
```diff
@@ -7,11 +7,19 @@ if [ $# -eq 0 ]; then
 fi
 NEWVER="$1"
 
+previous_channel() {
+  ruby -e "version = \"${NEWVER}\".match(/(\d+).(\d+).(\d+)/); \
+    minor = version[2].to_i - 1; \
+    puts \"channel/rubocop-#{version[1]}-#{minor}\""
```
maxjacobson: This is very clever. IMO it wouldn't be crazy to rewrite this whole script in ruby with a `#!/usr/bin/env ruby` shebang if that will make this all easier

Happy to stick with this, though. It might make your life a little easier to do something like:

```shell
cat <<EOM | ruby
  ruby goes here without need of escaping quotes
EOM
```

bin/newver
filipesperandio: hahaha struggled with this a bit too, never fully learned `ed`...
On master, this line is:
`gem "parser", "~> 2.4.0"`
But recent channels it is:
`gem "parser"`
The change make sure on either case the result is just `gem "parser"`... which could actually go away since it accomplishes nothing in the new approach for forking from latest channel... just realizing now that I am explaining it.

PR #196, opened 2019-06-05
Dog food

filipesperandio:


.rubocop.yml
```diff
@@ -1,6 +1,8 @@
 inherit_from: base_rubocop.yml
 
-require: rubocop-rspec
+require:
+  - rubocop-rspec
+  - rubocop-rails
```
wfleming: This isn't a rails app, so why do we want to apply the rails cops here? I wouldn't think we'd want to.

filipesperandio: Was using this to test the changes in the engine itself as the module is being removed from the core into an external module.

Also learned that we don't automatically trigger analysis for OSS repos on PRs made against non-default branches :(

wfleming: > Was using this to test the changes in the engine itself as the module is being removed from the core into an external module

That's not so much dogfooding IMHO as it is a strange way of doing QA. Since I don't think these rules are relevant to this engine, I don't think this proves anything other than that the module can be loaded, which could also be verified with a test or manual QA. If the rules aren't relevant to the source code here, I don't think we'd want to load it up for analysis runs.

> Also learned that we don't automatically trigger analysis for OSS repos on PRs made against non-default branches

That's not an OSS restriction, but a general limitation. OSS repos unfortunately do not have the workaround available to them that private repos do, which is adding the same repo multiple times with different default branches. https://docs.codeclimate.com/docs/code-climate-glossary#default-branch

filipesperandio:
> That's not so much dogfooding IMHO as it is a strange way of doing QA.

Yeah I mixed things up. Dog food would be having the branch using its own channel, but since it don't even trigger a build...

PR #198, opened 2019-06-06
Small text adjustments

marcelotoledo5000:
Just updates URL and Markdown syntax.


marcelotoledo5000:
Fixed, now is from master and to master again.

PR #201, opened 2019-06-11
Fix docs:scrape regex to skip forward to class definition

natesholland:
This is a bug that I originally found in #168 but we decided to pull out into its own PR since it is a separate thing than upgrading rubocop.

This scraper was missing the documentation on files like `lib/rubocop/cop/style/conditional_assignment.rb` and `lib/rubocop/cop/layout/indent_first_argument.rb` Because they had a comment before the main docs comment. This was causing them to be rejected because the docs were not long enough, despite the fact that these files do in fact have docs.

A solution I am proposing for this problem is to use a look ahead to grab the comments that come immediately before the class definition as this is where the docs comments always are.

Please let me know if this should go into a different branch, or if I should make PRs into multiple branches for this.

### DIFF
--- a/lib/tasks/docs.rake
+++ b/lib/tasks/docs.rake
@@ -18,7 +18,7 @@ namespace :docs do
       content = File.read(file)
       content = content.gsub(/.*\n\s+(?=module RuboCop)/, "")
 
-      class_doc = content.match(/(\s+#.*)+/).to_s
+      class_doc = content.match(/(\s+#.*)+(?=\n\s+class)/).to_s
       doc_lines = class_doc.
                   gsub(/^\n/, "").
                   gsub("@example", "### Example:").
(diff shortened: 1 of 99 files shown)


lib/tasks/docs.rake
```diff
@@ -18,7 +18,7 @@ namespace :docs do
       content = File.read(file)
       content = content.gsub(/.*\n\s+(?=module RuboCop)/, "")
 
-      class_doc = content.match(/(\s+#.*)+/).to_s
+      class_doc = content.match(/(\s+#.*)+(?=\n\s+class)/).to_s
```
filipesperandio: I don't think I fully understand this PR.
Is it fair to assume that this only change in the task impacted on all the `md` changes present here?

- I see a lot of stuff being dropped from the docs - are them all not wanted?
- I see some typos that make me think some where manually changed

I think we can address this against the latest channel (`rubocop-0-71` at this moment) which will then be include on the next channels going forward.

filipesperandio:
Closing for inactivity.

@natesholland If you wanna push this further, please re-open against the latest available channel, as of now `channel/rubocop-0-76`, and please, split the commits into what is a manual change and an automated one.

Thanks.

PR #207, opened 2019-08-12
Updated channel and documentation to Rubocop v0.74

filipesperandio:


lib/rubocop/config_patch.rb
```diff
@@ -10,4 +10,4 @@ def reject_obsolete_cops_and_parameters
   end
 end
 
-RuboCop::Config.prepend RuboCopConfigRescue
+RuboCop::ConfigObsoletion.prepend RuboCopConfigRescue
```
maxjacobson: :thinking: That's a curious name... any idea what it means?

filipesperandio: https://github.com/rubocop-hq/rubocop/blob/85f16b86ec79c13d3b2784e58d4d42f8944106e7/lib/rubocop/config_obsoletion.rb#L4

It was extracted from `Config` and first called `ObsoleteConfig`... seems to handle configs that get deprecated on old versions but my context is also minimum, meaning I learned about it just now. The spec for our patch started failing upon the upgrade.

PR #208, opened 2019-09-27
Add RubocopMinitest

duduribeiro:
Hey folks 👋

I don't know if this commit is valid. 

We use rubocop-minitest and it would be great if codeclimate could support it.

This commit includes rubocop-minitest gem in the Gemfile to allow
users that uses it to have their code analyzed with CodeClimate.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,8 +9,9 @@ gem "pry", require: false
 gem "rubocop", "0.75.1", require: false
 gem "rubocop-performance", require: false
 gem "rubocop-migrations", require: false
-gem "rubocop-rspec", require: false
+gem "rubocop-minitest", require: false
 gem "rubocop-rails", require: false
+gem "rubocop-rspec", require: false
 gem "safe_yaml"
 
 group :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -49,6 +49,8 @@ GEM
       unicode-display_width (>= 1.4.0, < 1.7)
     rubocop-migrations (0.1.2)
       rubocop (~> 0.41)
+    rubocop-minitest (0.4.1)
+      rubocop (>= 0.74)
     rubocop-performance (1.5.1)
       rubocop (>= 0.71.0)
     rubocop-rails (2.3.2)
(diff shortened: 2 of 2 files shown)


Gemfile.lock
```diff
@@ -47,6 +47,7 @@ GEM
       unicode-display_width (~> 1.0, >= 1.0.1)
     rubocop-migrations (0.1.2)
       rubocop (~> 0.41)
+    rubocop-minitest (0.0.1)
```
andyw8: The current release seems to be 0.2.1:

https://github.com/rubocop-hq/rubocop-minitest/releases

duduribeiro: hey @andyw8 👋 

Thanks for reviewing. Yes, current release is 0.2.1, but to use > 2.0, rubocop core should be updated. Can I update it?

thanks 🍻

andyw8:
I see, so that's causing it to resolve to an older version of rubocop-minitest.

There are 'channels' which use a newer version of RuboCop, e.g. https://github.com/codeclimate/codeclimate-rubocop/tree/channel/rubocop-0-74, so it might be better to create a PR against that branch.

duduribeiro:
Awesome!.. thanks for your suggestion

filipesperandio:
Sorry this is taking so long.
I am currently updating some channels to include latest rubocop version, if you don't mind doing this against the latest `channel/rubocop-0-76` branch I will be happy to merge this in.

duduribeiro:
hey @filipesperandio 👋 

thanks for the comment. I made it against the channel/rubocop-0-76 branch. Thanks so much 🍻

Gemfile
```diff
@@ -9,8 +9,10 @@ gem "pry", require: false
 gem "rubocop", "0.75.1", require: false
 gem "rubocop-performance", require: false
 gem "rubocop-migrations", require: false
-gem "rubocop-rspec", require: false
+gem "rubocop-minitest", require: false
+gem "rubocop-performance", require: false
```
filipesperandio: This got duplicated with the change.

duduribeiro: sorry for it.. pushed the fix. thanks!

PR #213, opened 2019-11-19
Support the thread_safety and i18n rubocop extensions.

imajes:
Adding in a couple more really useful rubocop extensions that would be nice to be default supported.


efueger:
Hi @imajes - thanks for the PR! I've added this to our list to review.

ale7714:
thanks @imajes !

PR #218, opened 2019-12-27
Pin bundler binary

filipesperandio:
Fix
```
      engine rubocop failed with status 1 and stderr
/usr/local/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require': cannot load such file -- rubocop (LoadError)
	from /usr/local/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
	from /usr/src/app/lib/cc/engine/rubocop.rb:6:in `<top (required)>'
	from /usr/local/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
	from /usr/local/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
	from /usr/src/app/bin/codeclimate-rubocop:8:in `<main>'
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN adduser -u 9000 -D app
 COPY Gemfile Gemfile.lock /usr/src/app/
 
 RUN apk add --update build-base git && \
-    gem install bundler && \
+    gem install bundler -v 2.0.2 && \
     bundle install --quiet -j 4 && \
     chown -R app:app /usr/local/bundle && \
     rm -fr ~/.gem ~/.bundle ~/.wh..gem && \


Dockerfile
```diff
@@ -10,7 +10,7 @@ RUN adduser -u 9000 -D app
 COPY Gemfile Gemfile.lock /usr/src/app/
 
 RUN apk add --update build-base git && \
-    gem install bundler && \
+    gem install bundler -v 2.0.2 && \
```
filipesperandio: Matches up https://github.com/codeclimate/codeclimate-rubocop/blob/channel/rubocop-0-76/Gemfile.lock#L96
We might want to eventually update the latest, but it will require some more digging to figure out what needs to be adjusted.

PR #221, opened 2020-01-31
Updated channel and documentation to Rubocop v0.78.0

filipesperandio:


lib/rubocop/config_patch.rb
```diff
@@ -8,6 +8,13 @@ def reject_obsolete_cops_and_parameters
...
+
+  def alert_about_unrecognized_cops(_)
+    super
+  rescue RuboCop::ValidationError => e
+    warn e.message
+  end
 end
 
 RuboCop::ConfigObsoletion.prepend RuboCopConfigRescue
+RuboCop::ConfigValidator.prepend RuboCopConfigRescue
```
filipesperandio: Had to add a patch so old cops don't raise, mostly because deprecated cops are being removed.

PR #225, opened 2020-02-25
Updated channel and documentation to Rubocop v0.80.0

DanielWright:
Most of this PR is pretty straightforward, and I'll leave some PR comments where I've got questions. At a minimum, this branch should provide a solid base for updating the channel.

### DIFF
--- a/config/contents/style/block_delimiters.md
+++ b/config/contents/style/block_delimiters.md
@@ -100,3 +100,29 @@ multi-line blocks.
     words.each { |word|
       word.flip.flop
     }
+
+### Example: BracesRequiredMethods: ['sig']
+
+    # Methods listed in the BracesRequiredMethods list, such as 'sig'
+    # in this example, will require `{...}` braces. This option takes
+    # precedence over all other configurations except IgnoredMethods.
+
+    # bad
+    sig do
+      params(
+        foo: string,
+      ).void
+    end
+    def bar(foo)
+      puts foo
+    end
+
+    # good
+    sig {
+      params(
+        foo: string,
+      ).void
+    }
+    def bar(foo)
+      puts foo
+    end
--- a/spec/cc/engine/config_upgrader_spec.rb
+++ b/spec/cc/engine/config_upgrader_spec.rb
@@ -18,11 +18,11 @@ module CC::Engine
         Style/FrozenStringLiteralComment:
           Enabled: false
         Style/HashEachMethods:
-          Enabled: false
+          Enabled: true
         Style/HashTransformKeys:
-          Enabled: false
+          Enabled: true
         Style/HashTransformValues:
-          Enabled: false
+          Enabled: true
       CONFIG
 
       # No warnings about obsolete cop name
--- a/spec/cc/engine/config_upgrader_spec.rb
+++ b/spec/cc/engine/config_upgrader_spec.rb
@@ -17,6 +17,12 @@ module CC::Engine
           Enabled: false
         Style/FrozenStringLiteralComment:
           Enabled: false
+        Style/HashEachMethods:
+          Enabled: false
+        Style/HashTransformKeys:
+          Enabled: false
+        Style/HashTransformValues:
+          Enabled: false
       CONFIG
 
       # No warnings about obsolete cop name
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,7 +2,10 @@ inherit_from: base_rubocop.yml
 
 require: rubocop-rspec
 
-Style/FileName:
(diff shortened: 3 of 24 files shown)


davehenton:
@filipesperandio

spec/cc/engine/config_upgrader_spec.rb
```diff
@@ -17,6 +17,12 @@ def get_true
           Enabled: false
         Style/FrozenStringLiteralComment:
           Enabled: false
+        Style/HashEachMethods:
+          Enabled: false
+        Style/HashTransformKeys:
+          Enabled: false
+        Style/HashTransformValues:
+          Enabled: false
```
DanielWright: While this change does cause the spec to pass, I'm not sure it's right way to handle the stdout warning about new cops.

DanielWright:
I should also note that the RSpec suite does run, but with some additional noise in the output from the following Rubocop warning:

```
The following cops were added to RuboCop, but are not configured. Please set Enabled to either `true` or `false` in your `.rubocop.yml` file:
 - Style/HashEachMethods
 - Style/HashTransformKeys
 - Style/HashTransformValues
```

andyw8:
@DanielWright that's expected due to https://github.com/rubocop-hq/rubocop/pull/7567

DanielWright:
Is it desirable to handle or suppress the warnings in this PR?

andyw8:
I guess the Code Climate team need to make a decision here on whether upgrading the channel should make the newer cops available by default or not.

spec/cc/engine/config_upgrader_spec.rb
filipesperandio: I would say let's turn them ON!

DanielWright: Your wish is my command! 51b5aa4

andyw8:
cc @pirj @bbatsov in case of any concerns about this.

bbatsov:
No concerns here. As state earlier, we switched to a policy where RuboCop will enable cops only at major releases and the rest of the time it'd be up to users to decide if they want to enable/disable some new cop. I view the message you mentioned as purely informational, not a warning. :-)

spec/cc/engine/config_upgrader_spec.rb
pean: Makes sense to enable them since they are enabled by default in Rubocop.

bbatsov: Note that they are not enabled by default in RuboCop, but rather "pending" which means the users have to decide whether to enable them or not. We'll likely enable them by default in version 1.0.

config/contents/style/block_delimiters.md
```diff
@@ -100,3 +100,29 @@ multi-line blocks.
...
+### Example: BracesRequiredMethods: ['sig']
+
+    # Methods listed in the BracesRequiredMethods list, such as 'sig'
+    # in this example, will require `{...}` braces. This option takes
+    # precedence over all other configurations except IgnoredMethods.
+
+    # bad
+    sig do
+      params(
+        foo: string,
```
pirj: Trailing comma, is this a deliberate choice?

spec/cc/engine/config_upgrader_spec.rb
filipesperandio: Ok, lets leave it on in the spec, but keeping it out of the default config, so it warns users to make a decision, just like rubocop itself.

filipesperandio: Ok, lets leave it on in the spec, but keeping it out of the default config, so it warns users to make a decision, just like rubocop itself.

config/contents/style/block_delimiters.md
DanielWright: This appears to be [a choice made upstream](https://github.com/rubocop-hq/rubocop/blob/v0.80.0/lib/rubocop/cop/style/block_delimiters.rb#L119), whether a typo or deliberate I cannot say.

spec/cc/engine/config_upgrader_spec.rb
DanielWright: Done in a81a64e. ✅

filipesperandio: Github is not helping much today... seems to be having issues.
I've tried to merge this early today and seems like the branch got into a weird state instead.

... just checked, it actually merged the changes in, but kept the PR open! So weird.

filipesperandio: The diff is now weird, but your changes are there.
I will be releasing this soon.

Issue #217, opened 2019-12-23
Support for rubocop-rake


andyw8:
[rubocop-rake](https://github.com/rubocop-hq/rubocop-rake) is another official plugin. It would be nice to have it supported.

More details:
https://metaredux.com/posts/2019/12/03/rubocop-meets-rake.html

filipesperandio:
It is becoming available starting on `channel/rubocop-0-83` onwards.

PR #228, opened 2020-03-20
Support rubocop-rake

pkopac:
Requested in #217
I could use it in our project as well.
When I include `rubocop-rake` in our `.rubocop.yml` `require` section, the rubocop step in codeclimate fails with:
```
/usr/local/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:117:in `require': cannot load such file -- rubocop-rake (LoadError)
...
```
Since this code runs separately from our codebase, I realized it doesn't matter what I include in our `Gemfile`. This gem must be included here.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,7 @@ gem "parser", "~> 2.4.0"
 gem "pry", require: false
 gem "rubocop", "~> 0.52.1", require: false
 gem "rubocop-migrations", require: false
+gem "rubocop-rake", require: false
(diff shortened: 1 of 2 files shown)


pkopac:
It would be great to release this in channel 0-80, do I have to open PR on that branch? Let me know ;)

filipesperandio:
@pkopac sorry this is being so late... please, do so against the latest channel branch, doing that we can keep it in the new channels moving forward.

PR #234, opened 2020-04-20
Updated channel and documentation to Rubocop v0.82.0

DanielWright:

PR #235, opened 2020-05-11
Updated channel and documentation to Rubocop v0.83.0

DanielWright:
This PR proposes to update the Rubocop integration to support v0.83.0. It includes commits from #234 as well, but I'm happy to rework the PR to do a straight upgrade from the 0-81 channel if that's preferable.


DanielWright:
Done and done!

PR #237, opened 2020-05-29
Updated channel and documentation to Rubocop v0.84.0

DanielWright:
This PR proposes to update the Rubocop integration to support v0.84.0.

PR #241, opened 2020-06-29
Updated channel and documentation to Rubocop v0.84.0

fede-moya:
Based on https://github.com/codeclimate/codeclimate-rubocop/pull/237


filipesperandio:
We can also rebase channel with 0-83 after we merge the `rubopcop-rake` so it gets here too.

fede-moya:
@filipesperandio Exactly Filipe, that was the idea. Will do now 👍

PR #256, opened 2020-09-09
Updated channel and documentation to Rubocop v0.90.0

DanielWright:
Better late than never, right?


noelia-lencina (approved):
Thanks for this!

PR #288, opened 2021-07-05
Add channel/rubocop-1-18-2

timlkelly:


fede-moya:
@timlkelly I would let's only rebase properly this one and we are done.

fede-moya:
@timlkelly or we could actually create a pr for version 18.3 actually

timlkelly:
@fede-moya Okay, I created #290. Can you make a 1.18.3 branch and then I will work through the rebase.

PR #290, opened 2021-07-06
Add channel/rubocop-1-18-3

timlkelly:


fede-moya (approved):
@timlkelly cool! Thank you 👏🏼

timlkelly:
Thanks for your help getting these merged! Are they available to use right away? Earlier I tried to add the rubocop-graphql gem to my project but it errored.

timlkelly:
Also, do the branches for the versions we skipped need to be removed because they're actually pointing to `1.10.0`

fede-moya:
> Thanks for your help getting these merged! Are they available to use right away?

fede-moya:
> Also, do the branches for the versions we skipped need to be removed because they're actually pointing to 1.10.0

Yeap, it's already on my TO-DO list.

timlkelly:
> > Thanks for your help getting these merged! Are they available to use right away?
> 
> Nop, the releasing process will take some time. I would advice you to check that again tomorrow.

@fede-moya has this been released? I'm still having difficulties with it and unsure if I'm missing a configuration step on my end. Or is there a place where I can see when it is available?

PR #319, opened 2022-06-22
QUA-619: Remove safe_yaml

dantevvp:
This PR removes the `safe_yaml` gem and uses `YAML.safe_load` instead to load up the `cops.yml` config file.

This fixes the `Style/WordArray` cop not raising issues when running analyses.

https://codeclimate.atlassian.net/browse/QUA-619

### DIFF
--- a/lib/cc/engine/issue.rb
+++ b/lib/cc/engine/issue.rb
@@ -1,9 +1,6 @@
 # frozen_string_literal: true
 
-require "safe_yaml"
 require "cc/engine/content_resolver"
-
-SafeYAML::OPTIONS[:default_mode] = :safe
 
 module CC
   module Engine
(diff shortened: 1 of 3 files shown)


lib/cc/engine/issue.rb
```diff
@@ -71,7 +68,12 @@ def cop_definition
       end
 
       def cop_list
-        @cop_list ||= YAML.load_file(expand_config_path("cops.yml"))
+        @cop_list ||= yaml_safe_load(expand_config_path("cops.yml"))
+      end
+
+      def yaml_safe_load(path)
+        yaml_code = ERB.new(File.read(path)).result
```
fede-moya, line 75: ⚠️  mmm, shouldn't this be a safe operation too ? ⚠️

fede-moya: isn't just `YAML.safe_load(File.read(path))` enough ?

f-moya (approved):
leaving behind a dependency sounds good enough on itself 🎉 🎉 , left a question...

BrianHawley:
Once the issues are resolved, can another PR be made to the channel/beta branch?

PR #326, opened 2022-07-26
QUA-619: remove mry gem and ConfigUpgrader class (rubocop-1-22-3)

dantevvp:
This PR removes the `mry` gem and the `ConfigUpgrader` class, that was using that gem to migrate older rubocop configuration attributes to their newer equivalents.

The `mry` gem is intended to be used in pre-1.0 versions of rubocop. In more recent versions, this gem seems to behave unpredictably, breaking some specific cops, so there's a reason to remove it entirely for rubocop versions post 1.0

The `ConfigUpgrader` class was removed because it's sole purpose was to use `mry` to migrate config files.

Example:

This is the CLI analyzing a piece of code with rubocop version 1.22.3, while using `mry` to migrate configs

image

When removing the mry gem, running the same analysis on the same piece of code returns this:

image


f-moya (commented):
Hey @dantevvp, It looks good. Awesome finding. Could you also attach captures of the actual expected output by just running rubocop without codeclimate to confirm that this will indeed induce correct results, please ?

Also, It's worth noticing that this might increase users's existent issues without them actually changing a line of code. So I think I would consider to include this on newer channels, and also running some tests across different repos to grasp a sense of how much more issues can this change introduce on a given codebase.

You might want to create a dedicated channel for this kind of testing.

fede-moya:
@larkinscott Looping you in on this one, I believe you showed interest on it 😄

larkinscott:
+1 on this being a great finding!

> Also, It's worth noticing that this might increase users's existent issues without them actually changing a line of code. So I think I would consider to include this on newer channels, and also running some tests across different repos to grasp a sense of how much more issues can this change introduce on a given codebase.

This is definitely a good idea.

dantevvp:
@fede-moya Ok, if I understand correctly the following steps would be:

- Close this PR
- Open a new one with this change to merge on `channel/beta`
- Make sure that newer channels include this change

I'll also try to make a list of the cops that `mry` breaks. I remember being able to see that while testing.

Is this correct?

fede-moya:
@dantevvp Yes that's correct, except that before pushing in to the `beta` channel I would deploy it into a `channel/test` branch so we can test it against different repos online to measure how more issues could this create on a given repo, unless you have already done using the CLI.

PR #328, opened 2022-08-08
QUA-619: remove mry gem (beta)

dantevvp:
This PR removes the `mry` gem and the `ConfigUpgrader` class, that was using that gem to migrate older rubocop configuration attributes to their newer equivalents.

The `mry` gem is intended to be used in pre-1.0 versions of rubocop. In more recent versions, this gem seems to behave unpredictably, breaking some specific cops, so there's a reason to remove it entirely for rubocop versions post 1.0

The `ConfigUpgrader` class was removed because it's sole purpose was to use `mry` to migrate config files.

Example:

This is the CLI analyzing a piece of code with rubocop version 1.22.3, while using `mry` to migrate configs

image

When removing the mry gem, running the same analysis on the same piece of code returns this:

image

NOTE: the removal of this gem makes the engine produce more issues compared to previous versions in some cases. There might be a big difference in new issues that weren't being caught before when using this update (between 10-100 in large codebases).


BrianHawley:
You should also remove spec/support/config_upgrader_rubocop.yml because it's only used in that deleted test.

f-moya (approved):

PR #339, opened 2022-12-16
Fix CircleCI configuration

BrianHawley:
CircleCI sent me an email indicating that the config settings for this repo were set to use a deprecated image (Ubuntu 14.04) because we didn't specify the image to use.

The image option is required, per their docs; set it as recommended.


BrianHawley:
@fede-moya or @filipesperandio could you review and merge?

This should also be able to be cherry-picked to other new channels, though they really should be based on channel/beta anyways so that shouldn't be a problem. The particular setting only affects the process of creating a new version of the engine image in docker, and only affects the host process that is building the image. It doesn't affect the built image itself, which is based on Dockerfile instead of this config.

BrianHawley:
@fede-moya this change is still needed. Could you review?

PR #343, opened 2023-02-13
Push images to Dockerhub instead of GCR (rubocop-1-39-0)

dantevvp:
Deprecate pushing of the image to GCR in favor of pushing to Dockerhub.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,11 @@
-.PHONY: image test docs bundle
+.PHONY: image test docs bundle release
 
 IMAGE_NAME ?= codeclimate/codeclimate-rubocop
+RELEASE_REGISTRY ?= codeclimate
+
+ifndef RELEASE_TAG
+override RELEASE_TAG = latest
+endif
 
 image:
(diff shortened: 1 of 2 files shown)


RubyBrewsday (approved):

Makefile
```diff
@@ -1,6 +1,11 @@
-.PHONY: image test docs bundle
+.PHONY: image test docs bundle release
 
 IMAGE_NAME ?= codeclimate/codeclimate-rubocop
+RELEASE_REGISTRY ?= codeclimate
+
+ifndef RELEASE_TAG
+override RELEASE_TAG = latest
+endif
```
fede-moya, line 8:
```suggestion
RELEASE_TAG ?= latest
```

dantevvp: @fede-moya this sadly does not work when providing an empty string. Makefile takes empty strings as not null, so it'll end up not tagging the image as `latest`. The `ifndef` workaround lets us provide an empty string that will still be replaced with `latest`

PR #349, opened 2023-03-17
Channel/rubocop 1 48 1

SOyeniyi801:


BrianHawley:
This won't work: channel/beta has fixes that 1.39 didn't have. It's better to base channel/rubocop-1-48-1 off of channel/beta after #350 is merged.

PR #353, opened 2023-04-11
allow standard to be run with rubocop. Fixes #351

### LINKED ISSUE #351: Add support for standardrb
We use [standard](https://github.com/testdouble/standard) along with rubocop in our project. We use the `inherit_gem` config in `.rubocop.yml` to add on to standard's config. We also use the `require` config to require standard along with other gems we use (`rubocop-performance`, `rubocop-rails`, etc).

Can a channel be made that requires `standard`?

RubyBrewsday:


fede-moya (approved):
Looks good!

PR #354, opened 2023-04-17
add `syntax_tree` gem for rubocop-1-18-3

RubyBrewsday:
* Repos that use rubocop-1-18-3 in analysis can now integrate `syntax_tree` into their rubocop configurations:
https://github.com/ruby-syntax-tree/syntax_tree#rubocop

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -20,6 +20,7 @@ GEM
     parallel (1.20.1)
     parser (3.0.1.1)
       ast (~> 2.4.1)
+    prettier_print (1.2.1)
     pry (0.14.1)
       coderay (~> 1.1)
       method_source (~> 1.0)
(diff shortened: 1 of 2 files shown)


larkinscott (approved):

Gemfile.lock
```diff
@@ -20,6 +20,7 @@ GEM
     parallel (1.20.1)
     parser (3.0.1.1)
       ast (~> 2.4.1)
+    prettier_print (1.2.1)
```
mlinderman, line 23: curious about why this upgrade to prettier_print when 1.2.0 is required by syntax_tree...

RubyBrewsday: Ah so this was automatically set when I added in the `syntax_tree` gem and ran `bundle install`. Since `syntax_tree` for version `6.1.1` has `>= 1.2.0` set as a dependency, I'm assuming that `1.2.1` is the highest version that meets that requirement.

mlinderman (approved):

PR #367, opened 2024-04-04
Add rubocop-rspec_rails to channel/rubocop-1-56-3

fonji:
Rubocop-rspec has extracted the rails cops to rubocop-rspec_rails
> the cops related to rspec-rails are aliased (RSpec/Rails/Foo == RSpecRails/Foo) until v3.0 is released

https://github.com/rubocop/rubocop-rspec/blob/master/CHANGELOG.md#2280-2024-03-30

So it would be nice to support it before v3 is released

PR #368, opened 2024-04-04
Add rubocop-rspec_rails to beta

fonji:
Same thing as #367 for the beta channel

Rubocop-rspec has extracted the rails cops to rubocop-rspec_rails
> the cops related to rspec-rails are aliased (RSpec/Rails/Foo == RSpecRails/Foo) until v3.0 is released

https://github.com/rubocop/rubocop-rspec/blob/master/CHANGELOG.md#2280-2024-03-30

So it would be nice to support it before v3 is released

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,6 +13,7 @@ gem "rubocop-performance", require: false
 gem "rubocop-rails", require: false
 gem "rubocop-rake", require: false
 gem "rubocop-rspec", require: false
+gem "rubocop-rspec_rails", require: false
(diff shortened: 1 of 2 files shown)


hallelujah:
@fede-moya any chance this could be merged?

vietqhoang:
+1 in getting this into beta. At the moment this particular issue is breaking our CodeClimate integration. We would like to not have to remove departments from the configuration. Open to the option to have CodeClimate ignore these departments, if there is one.

PR #374, opened 2024-11-09
Update rubocop to 1.75.2

BrianHawley:
- General bundle update.
- Update the docs too.
- Fix rubocop API deprecations.
- Fix reasonable rubocop complaints.
- Disable unnecessary rubocop complaints.
- Explicit ostruct dependency for future compatibility.
- Remove an unnecessary empty file.


BrianHawley:
@f-moya could you merge this one? It's has a bundle update and rake docs:scrape, but it also fixes the rubocop failures in the code itself, and all of the new rubocop API deprecation warnings in the new version.

f-moya:
Hi @BrianHawley, I hope you're doing well. I am no longer with Code Climate, so I won't be able to assist with the deployments moving forward.

BrianHawley:
@f-moya do you know whose job it is at Code Climate to review and merge these PRs? So I can tag them when I make these updates.

sbsurf (approved):

kstevens715:
Getting this released would be a big help to me. I'm using ruby-lsp and Neovim and it sounds like this should fix the issue according to: https://github.com/Shopify/ruby-lsp/issues/2840#issuecomment-2468010949

gfguthrie:
@BrianHawley `standard` 1.43.0 supports `rubocop` 1.69.1 – any thoughts on adding it back in for this channel? Would you accept a PR to your PR if I figured it out?

BrianHawley:
@gfguthrie I checked with the standard people and they were pretty clear that they have no intention to keep their development in sync with that of rubocop, as it's (understandably) too much work. So if it's temporarily in sync, we can't assume that this will continue. It would be a bad precedent to add standard back when we'd need to remove it again in the near future.

It still is the best approach to have a separate set of standard channels, with a separate beta channel that would have the latest standard and the latest rubocop that standard supports. As I mentioned before in the related issue.

kbrock:
rubocop is up to 0.93.1
Getting a channel for a recent version of rubocop would be great

BrianHawley:
@kbrock it's updated. But I still don't know which GitHub user account to tag to get these PRs merged. The previous person no longer has merge privileges, as mentioned above.

